PR #1, opened 2015-04-29
Add Data from Denmark

tmtmtmtm:
Take data from https://denmark.popit.mysociety.org/
- remove people who don't have `currentMP` set
- Expand the party abbreviations to full names
- Add dates of the current term


briatte:
The current legislature of the Folketing is at

http://www.ft.dk/Folketinget/searchResults.aspx?letter=ALLE&pageSize=500&pageNr=

[This scraper](https://github.com/briatte/folketinget/blob/master/data.r) also scrapes MPs from previous legislatures, based on the links to their profiles found by scraping private bills.

The links use letters, as in

http://www.ft.dk/Folketinget/findMedlem/SBEEN.aspx

tmtmtmtm:
After the 2015 election I updated this with the new information: https://github.com/everypolitician/everypolitician-data/pull/341/

But we still need to back and match these with the historic records.

http://www.hvemstemmerhvad.dk/api/api.php also provides a lot of useful information, although I couldn't see a way to discover _when_ everyone was a member from there. 

http://aarhus13.jboye.com/speaker/jorn-sjostrom/ also states (in 2013) that they were working on an official Parliament Open Data platform, so I keep hoping that I won't actually _need_ to scrape historic info!

PR #3, opened 2015-04-29
Italy

tmtmtmtm:
Create data for Italy from https://parlamento-test.popit.mysociety.org/

In passing, set up an easy way to set data for the current term, and switch Denmark and Wales to also use it.

### DIFF
--- a/denmark/Rakefile.rb
+++ b/denmark/Rakefile.rb
@@ -5,6 +5,12 @@ require_relative '../rakefile_common.rb'
 
 @DEST = 'denmark'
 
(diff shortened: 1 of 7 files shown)


briatte:
I've written scrapers that also cover [legislatures/terms 13 to 16](https://github.com/briatte/parlamento), which are a complete mess; [this script](https://github.com/briatte/parlamento/blob/master/data-ca-old.r), for instance, shows how the syntax changes between legislatures 13, 14 and 15. The scraper covers both chambers.

PR #6, opened 2015-04-30
Bring in data from parldata.eu

duncanparkes:
Adds a small python script to collect together the data we need from parldata.eu's API and make it into some JSON files.

At the moment this is just collecting the data rather than massaging it into formats we like.


tmtmtmtm:
We probably want (/need) to credit somewhere that the vpapi script isn't ours, but comes from KohoVolit/api.parldata.eu

tmtmtmtm:
Rather than a single script that knows about every legislature, and writes them out to named files, this should be a standalone unix-style script that takes as an argument `al/kuvendi` (or whatever), and just outputs to STDOUT. 

Then the individual Rakefiles can call it with the relevant parliament identifier, and handle its output whatever way they want.

I tried doing this myself, but I don't even know the incantation to run this, as a plain `python parldataeu.py` fails with 

```
Traceback (most recent call last):
  File "parldataeu.py", line 5, in <module>
    import vpapi
  File "/Users/tony/Code/everypolitician-data/vpapi.py", line 5, in <module>
    import requests
ImportError: No module named requests
```

duncanparkes:
That's because you don't have the requests library installed. I guess we should make a virtualenv and then install it in there.

I totally agree it should be a standalone script, but I suggest we do that to it the next time parldata.eu add a dataset.

tmtmtmtm:
Having an entirely different pipeline for creating data for from this source is going to be much more complicated and confusing than keeping the process as similar as possible for all data.

But I'll merge this, and migrate towards that myself.

parldataeu.py
```diff
@@ -0,0 +1,110 @@
...
+import json
+
+import vpapi
+
+parliaments = [
+    ('al/kuvendi', 'albania', 'kuvendi'),
+    ('cz/psp', 'czech_republic', 'psp'),
+    ('cz/senat', 'czech_republic', 'senat'),
+    ('hu/orszaggyules', 'hungary', 'orszaggyules'),
+    ('me/skupstina', 'macedonia', 'skupstina'),
```
tmtmtmtm: Missed this first time around, but `me` is Montenegro, not Macedonia (`mk`)

PR #18, opened 2015-05-04
Add Iceland

tmtmtmtm:
Add data for Iceland via https://morph.io/tmtmtmtm/iceland-althing-wp


briatte:
[Here's my scraper for Iceland](https://github.com/briatte/althing/blob/master/data.r). I've used online profile pages and CV pages to sort out stuff like time in office. There's quite a few party switches during legislatures.

PR #148, opened 2015-05-25
Finland: initial data

### LINKED ISSUE #46: Include Finland
Finland has been grandfathered into the live data, but doesn't exist here.

tmtmtmtm:
We had grandfathered this in — load it properly now.

Closes #46

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -52,3 +52,16 @@ task :publish do
   end
 end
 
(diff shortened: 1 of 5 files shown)


briatte:
My totally messy scraper for Finland is [there](https://github.com/briatte/eduskunta/blob/master/data.r). Nothing to learn from it, but I'll just reference it.

PR #222, opened 2015-06-09
Add Czech Republic

tmtmtmtm:


briatte:
Referencing [my own scraper](https://github.com/briatte/poslanecka/blob/master/data.r) in case it helps (I'm scraping private bill sponsors only, using links from the bills). Goes back to legislature 2 (1996).

briatte:
Update: I've written a [different scraper](https://github.com/briatte/parlament/blob/master/data-se.r) for the Czech Senate.

The scraper first visits all [constituency pages](http://senat.cz/senat/volby/hledani/obvody.php?ke_dni=28.10.2015&O=10) using the URL `http://senat.cz/senat/volby/hledani/o_obvodu.php?kod=ID`, where `ID` goes from `1` to `81`. Each constituency page has links to senator pages in `.tHistory td a`.

Note that Czech senators are elected on six-year mandates through a continuous system: one third of the constituencies gets renewed every two years. Consequently, you won't be able to match legislatures 2-7 of the lower chamber: instead, you will get 10 "legislatures" from 1996 to today. Same-year elections are not even synchronous (e.g. the 1996 Senate election was not organised on the same date as the lower chamber election that same year).

PR #282, opened 2015-06-25
Add a `slug` for each country and chamber

tmtmtmtm:
The `name` of the Legislature should hopefully become the fuller version (e.g “House of Representatives” rather than just “House” RSN, so we should have a `slug` that can be used in URLs etc that is hopefully more unchanging.

The general consensus is that this should use “-“ rather than “_” as a separator (because Google, largely)


chrismytton:
:+1:

PR #309, opened 2015-06-30
Ensure no Rakefiles supply TERMS

tmtmtmtm:
If we’re providing any term meta-info ourselves, it should be in a terms.csv file, not embedded in the Rakefile


tmtmtmtm:
@chrismytton — here's another example of one that took a very long time to trigger…

PR #322, opened 2015-07-02
Add Hungary

tmtmtmtm:


briatte:
Referencing [my own scraper](https://github.com/briatte/orszaggyules/blob/master/data.r), which covers less legislatures (only scraping MPs for which private bills are also available).

PR #364, opened 2015-07-11
France: Initial data

### LINKED ISSUE #66: France
http://cpc.regardscitoyens.org/trac/wiki/API

Getting current data is easy. There appears to be a call for previous terms, but you need to know the value to give.

tmtmtmtm:
Closes #66


briatte:
I'm failing to find the code to the actual scraper (for France or for any other country). Am I missing something? I'm interested because of [this project](https://github.com/briatte/parlnet), which also scrapes MP details.

tmtmtmtm:
The scrapers usually live elsewhere, and the data is fetched into here and post-processed. You can usually find the details on where the data comes from in the `sources/` directory (in most cases in `sources/morph`). In this case the relevant file is https://github.com/everypolitician/everypolitician-data/blob/02d74367704cb67142dd8fc3b4cb6cf02125aacf/data/France/National_Assembly/sources/morph/instructions.json, which points at https://morph.io/tmtmtmtm/nos_deputes_france, which runs https://github.com/tmtmtmtm/nos_deputes_france/blob/master/scraper.rb to gather the data from the NosDeputes.fr API (http://www.nosdeputes.fr/deputes/json)

About 90% of the scrapers involve getting the data from Parliament websites — in this case we were able to ride on the back of the work that NosDeputes had already done.

I'd love to hear more about what you're doing with parlnet, and we should definitely talk further about ways we might be able to collaborate! I'm tony@mysociety.org if you want to go to email…

briatte:
Thanks, that's very helpful!

Several of the scrapers make good use of Wikidata. I wish I had known about it earlier :)

Re: `parlnet`, I'm building networks for [visualization](http://f.briatte.org/parlviz/) and analysis purposes. The scrapers are pretty dirty, but they work for what I need to do (which involves modelling the networks to measure e.g. the extent of partisanship in a given chamber).

tmtmtmtm:
The Wikidata information is a little hit and miss at the minute, and the tooling around it isn't great yet either, but I've started abstracting some of our recurring patterns out into https://github.com/everypolitician/wikidata-fetcher, which you can see in action on things like https://morph.io/tmtmtmtm/france-chamber-of-deputies-wikidata (from https://github.com/tmtmtmtm/france-chamber-of-deputies-wikidata/blob/master/scraper.rb)

At some point soon I'd like to work out how to start feeding a lot more information back into Wikidata too.

Will contact you separately about `parlnet`.

PR #373, opened 2015-07-15
Poland: Put data live

tmtmtmtm:
Problems with Parldata information:

* no dates on 'chamber'
* no dates on memberships of legislature
* no dates on memberships of Party (and can have been in more than one in the period)
* no area information


tmtmtmtm:
Replaced by #489

PR #375, opened 2015-07-15
Slovakia: Initial data

tmtmtmtm:


briatte:
If it helps, [this scraper](https://github.com/briatte/nrsr/blob/master/data.r) covers legislatures/terms 1 to 6.

The MPs are there:

http://www.nrsr.sk/web/Default.aspx?sid=poslanci%2fzmeny

This is the list of 'seat changes', not the list of MPs in a given legislature. This list is more complete than the term-specific lists.

My own scraper is pretty messy because of difficulties with ASP/doPostBack.

Issue #141, opened 2015-05-24
Solomon Islands


tmtmtmtm:
http://www.parliament.gov.sb/index.php?q=node/134 (10 terms)

tmtmtmtm:
No parties listed on main pages, but a list of members by party is at http://www.parliament.gov.sb/index.php?q=node/147

tmtmtmtm:
The list of members by party only includes historic members — not the current ones. @rubeenam, can you see if you can find party memberships somewhere?

rubeenam:
Sure, I am looking!

rubeenam:
MPs from 2014 general elections, parties and constituencies

http://www.siec.gov.sb/index.php/journalist/127-2014-national-general-election-results

tmtmtmtm:
Excellent, thanks! I'll try merging that in later today.

PR #404, opened 2015-07-23
Solomon Islands: Initial data

tmtmtmtm:
part of #141


tmtmtmtm:
@chrismytton : any idea why this hasn't triggered a parallel PR on viewer-sinatra?

chrismytton:
@tmtmtmtm Looks like the last 2 webhook deliveries timed out for some reason.

screen shot 2015-07-23 at 10 22 25

(that's the one for labeling the pull request, but the opened one timed out as well)

I've redelivered the webhook and it seems to have worked now so I guess it was a transient issue.

chrismytton:
Hmm, it now seems that the deployment job is failing with the following error:

```
Errno::ETIMEDOUT: Connection timed out - connect(2) for "api.github.com" port 443
```

So there seems to be some connectivity issues between Heroku's EU region and GitHub at the moment. The job has been queued for a retry so should Just Work once connectivity is fixed. I can manually retry it in a few hours if that's not the case.

chrismytton:
Ah, as soon as I posted that it finished successfully!

tmtmtmtm:
yep, all seems to be working now. Thanks!

PR #735, opened 2015-09-08
Fix quote in readme

henare:
Remove smart quote and replace it with a normal (dumb?) one. Makes it
so you can copy and past working code.


tmtmtmtm:
Thanks @henare

PR #833, opened 2015-09-13
Gibraltar: initial data

### LINKED ISSUE #284: Gibraltar
http://www.parliament.gi/history/composition-of-parliament — lists everyone back to the First House of Assembly in 1969

Looks quite awkward to scrape though — would likely be much easier done by making a spreadsheet by hand.

andylolz:
Scraper at https://morph.io/andylolz/gibraltar-parliament

~~Not sure I’ve got this quite right… Some politicians are clearly duplicates, but appear with multiple IDs. Let me know if I can do more to fix this!~~ UPDATE: this should be okay now.

Closes #284.


tmtmtmtm:
Thanks for this @andylolz. Now staged at https://everypolitician-viewer-pr-658.herokuapp.com/gibraltar/

It's looking mostly good, but a few tweaks will make it even better:

Firstly, the dates in the term tables will need to be YYYY-MM-DD or YYYY style (doc patches welcomed too!) so these sort properly (I should probably add something into the merge process to error on that, or maybe even try to convert them).

Secondly, we'll need to handle the change in the body itself, from Assembly to Parliament. It's not entirely clear to me, however, whether this was purely a change of name, or actually becoming a different entity.

I would normally assume the latter, but keeping the numbering going, from 10th Assembly to 11th Parliament suggests it might just have been the former. (I note that Wikipedia act more like it was the latter.)

If it *was* the latter, then we can easily chop that up in the config, and split everything across two different subdirectories (and a `WHERE term > X` and `WHERE term < X` in the relevant `query` entries of each `instructions.json`), but if it's the former then you should really treat both "10"s as the same thing, and we'll record elsewhere the date for the change of name.

To help resolve duplicate people, I'd suggest splitting off the honorific prefixes and suffixes into separate columns. This isn't completely necessary — a lot of data on the site includes them — but it'll stop "A P MONTEGRIFFO OBE" becoming a different `Person` from `A P MONTEGRIFFO, O.B.E.`, or "R R RHODA QC" from "R R Rhoda, QC" etc (including the 4 different variations of E M BRITTO). You'd probably also need to standardise how initials appear too, to help with "Major R J PELIZA" vs "Major R. J. PELIZA"

It would also be good to get at least some of the Party affiliations. https://en.wikipedia.org/wiki/Next_Gibraltar_general_election lists these for the current members and is trivially scrapable — I haven't yet investigated whether there are suitable pages for previous ones. I'm happy to do that myself, but if you would prefer to do this yourself I'm happy with that to :)  For things like that I generally prefer a second scraper, and then merging the data together in the `instructions.json` (e.g. [Costa Rica](https://github.com/everypolitician/everypolitician-data/blob/master/data/Costa_Rica/Assembly/sources/instructions.json)). If you _do_ want to do that yourself, you should also store the `<a title="Whatever">` version of people's names too, as that will allow us to look them up in Wikidata and create Wikipedia URLs for them (though you should take care to omit incipient links: `path('.//a[not(@class="new")]/@title'` works in XPath)

andylolz:
Okay – I’ve fixed dates, honorifics, and matched people by their names. It looks okay.

I’ve assumed the change from House of Assembly to Parliament was just a name change. From [Gibraltar Parliament (wikipedia)](https://en.wikipedia.org/wiki/Gibraltar_Parliament):
> Between 1969 and 2006 it was called the Gibraltar House of Assembly.

…also [Gibraltar Constitution Order 2006 (wikipedia)](https://en.wikipedia.org/wiki/Gibraltar_Constitution_Order_2006#Contents):
> Among changes introduced by the new constitution were:
> 
>  * Renaming the House of Assembly to the Gibraltar Parliament.

However, wikipedia does appear to number from 1 again (i.e. they refer to the 12th parliament as the 2nd).

Is it best to force push updates to this branch, or add new commits, or leave it to you to do in #835?

andylolz:
I’ve added a scraper for that wikipedia data just here: https://morph.io/andylolz/gibraltar-wiki-current but I’m not exactly sure how to combine them. My instructions.json file isn’t working currently!

tmtmtmtm:
Thanks @andylolz . Sorry for the less-than-helpful error message!

Could you also grab the 'title' tag for people's wikinames? That's much easier to cross-match with than the URL version.

andylolz:
No problem! I’ve _almost_ got it now.

> Could you also grab the 'title' tag for people's wikinames?

The `name` field [in the wiki-current scraper](https://morph.io/andylolz/gibraltar-wiki-current) is from the link title tag. Is that what you mean?

tmtmtmtm:
Ah, OK — great! I assumed that was the displayed version (but where are you storing that, then?)

andylolz:
The displayed version (the anchor text) is stored in `sort_name`

tmtmtmtm:
It's a little odd, as usually you'd want sort to go by surname first…

tmtmtmtm:
I've also added https://morph.io/tmtmtmtm/gibraltar-parliament-wikidata, which you might want to merge in too :)

andylolz:
> It's a little odd, as usually you'd want sort to go by surname first…

Wikipedia lists them surname first: https://en.wikipedia.org/wiki/Next_Gibraltar_general_election#Current_membership

So I’ve put the anchor text (e.g. “Balban, Paul John”) in `sort_name`, and the title text (e.g. “Paul Balban”) in `name`.

> I've also added https://morph.io/tmtmtmtm/gibraltar-parliament-wikidata, which you might want to merge in too :)

Great! I’ll try adding that now.

andylolz:
Ahh – merging the wikidata stuff has helped me understand what you meant about the wikipedia name now! I‘ve added a `wikipedia_name` column to the [gibraltar-wiki-current scraper](https://morph.io/andylolz/gibraltar-wiki-current) for that, in the same way the Australia importer does.

tmtmtmtm:
Great — I've updated the Wikidata scraper for the new location and column name. You think this is ready to go now?

andylolz:
I changed the wikipedia scraper to look at historic elections, too (it now runs from [this category page](https://en.wikipedia.org/wiki/Category:Elections_in_Gibraltar) and looks at every general election page).

I’ve pushed the `amatch_threshold` down to 40 which is perhaps a bit too low… But it means that e.g. “Edwin Joseph Reyes” matches “E J Reyes”.

tmtmtmtm:
Yay. 

I generally tweak the thresholds until it matches everyone it should, and no-one it doesn't, in the current data, and then set the warning threshold to something that should make it fairly obvious if someone new gets picked up incorrectly later.

data/Gibraltar/Parliament/sources/instructions.json
```diff
@@ -0,0 +1,59 @@
...
+      "file": "morph/terms.csv",
+      "create": {
+        "file": "morph/terms.csv",
+        "type": "morph",
+        "scraper": "andylolz/gibraltar-parliament",
+        "query": "SELECT * FROM terms"
+      },
+      "source": "http://www.parliament.gi/history/composition-of-parliament",
+      "type": "term",
+      "merge": {
```
tmtmtmtm: At the minute there's no way to merge 'term' data — the code will only work with a single 'terms' source. This section will just be ignored, rather than causing any trouble, but it's probably better to remove it just in case something changes later, and this starts doing something unexpected.

data/Gibraltar/Parliament/sources/instructions.json
```diff
@@ -0,0 +1,59 @@
...
+        "scraper": "andylolz/gibraltar-parliament",
+        "query": "SELECT * FROM data"
+      },
+      "source": "http://www.parliament.gi/history/composition-of-parliament",
+      "type": "membership"
+    },
+    {
+      "file": "morph/terms.csv",
+      "create": {
+        "file": "morph/terms.csv",
```
tmtmtmtm: there's no need to duplicate this line to here — the `create` destination will always be the `file` (I expect you just copied this from somewhere else that also makes the same mistake, but you may as well remove it)

tmtmtmtm:
Couple of tiny nits about the instructions file, but otherwise this is pretty much ready to go.

Having some people in ALLCAPS and some in 'normal' case (e.g. at https://everypolitician-viewer-pr-658.herokuapp.com/gibraltar/parliament/term-table/12.html) is a little odd, though. I'm not quite sure why that's happening, as later sources [should only be adding fields that don't exist from earlier sources](https://github.com/everypolitician/everypolitician-data/blob/b02740a2caf89304f222bbcbe70471f7a7767b13/rakefile_merged.rb#L131), so the ones from the 'gibraltar-parliament' source should "win". I'll have a look at that later.

andylolz:
> I generally tweak the thresholds until it matches everyone it should, and no-one it doesn't

Kk cool, that’s what I went for too. There are no obvious errors at the moment.

> later sources should only be adding fields that don't exist from earlier sources, so the ones from the 'gibraltar-parliament' source should "win"

Ahh, right. The gibraltar-parliament ones are indeed winning, but that scraper [outputs names differently depending on the term](https://github.com/andylolz/everypolitician-data/blob/3abb33ec/data/Gibraltar/Parliament/sources/morph/data.csv#L153-L158) (because of the source data.) I guess I could titlecase names, but I was avoiding it!

tmtmtmtm:
https://github.com/openaustralia/morph/issues/905 — but looks like it's back now!

andylolz:
Right – latest issue (sorry this is taking forever!)

In the [gibraltar-wikipedia scraper](https://morph.io/andylolz/gibraltar-wikipedia), I don’t currently state for which terms a politician is affiliated with a particular party. The instruction in question has type `person`. There’s no error, and the importer does import something, but it’s a bit of a muddle. e.g. Jay Bossano should be GSLP for all 11 of his terms, but instead [he’s sometimes affiliated with `party/unknown`](https://github.com/andylolz/everypolitician-data/blob/80cb30e1/data/Gibraltar/Parliament/ep-popolo-v1.0.json#L1248-L1345).

I wonder if I need two separate instructions here: One `person` instruction to import the wikipedia data. Then a separate `membership` instruction to import party membership for a particular person & term. The reason to do this in two steps is it doesn’t look like `membership`s can be merged with an `amatch_threshold` (it gets a “not implemented yet” error.)

andylolz:
Deleted my previous comment, because it was v confusing!

I think the solution here is to combine the gibraltar-wikipedia scraper and gibraltar-parliament scraper into a single scraper. It doesn’t look like `rakefile_merged.rb` can merge two sets of `membership` data on two fields (`name` and `term`) with a fuzzymatch on `name`.

tmtmtmtm:
@andylolz — I'd rather work out what was wrong with merging them and fix that, as that will be better long term than having to combine the scrapers. (I haven't had a chance to look at what the issue was yet, as I'm not officially here today :))

tmtmtmtm:
@andylolz is there actual new membership information (i.e. simply that someone was a representative in a given term?) in the Wikipedia scraper that's not in the Parliament one? I had thought that the Wikipedia scraper was mostly just for getting Party information? I don't think I'll get a chance to look at this in any detail tonight, but I'm happy to spend some time tomorrow getting to the bottom of it all.

andylolz:
> Sorry you're getting caught up slightly in the middle of that.

No problem! I’ve looked at the `rakefile_merged.rb` code now, and it’s clear that it has evolved that way (not in a bad way! But there are various “Not implemented yet” bits).

Yeah – I think I agree that it’s better to fix the merging. I don’t think my requirement here is _that_ weird :)

So… [This bit of rakefile_merged.rb](https://github.com/everypolitician/everypolitician-data/blob/9991db85/rakefile_merged.rb#L176-L177) performs a match based on a field (in my case, `name`). The haystack is `all_rows`. The `term` is totally ignored. I need either of the following to happen instead:

 * FuzzyMatch should also take into account the `term` field. This is something like the `groupings` option that FuzzyMatch has, but I couldn’t figure out how to do a grouping over a different field
 * the haystack that FuzzyMatch uses should only be rows with a matching `term`. (NB this is what the `groupings` thing does anyway)

tmtmtmtm:
> it’s clear that it has evolved that way (not in a bad way! But there are various “Not implemented yet” bits).

Oh, it's even worse than that :)

Some of it is left-behind remnants of approaches that turned out to be dead ends — and I think the part you're looking at might be that. I don't think you actually want a Membership merge at all: that's really for the case where you have a separate scraper for Term 9 than from Term 10, etc.  If you're really just adding extra information to existing data (i.e. adding new columns to existing rows, rather than new rows) then you should be using the Person merge, which is the part that's gradually being factored out into the CSV-Merge class further up. 

If it _is_ the case that you're adding both new rows _and_ new columns from the Wikipedia scraper, then yeah, the code doesn't really try to cope with that yet, as it hasn't yet needed to. I have a reasonably clear idea in my head of what the ideal interface for this _could_ end up like, but I'd rather evolve towards it based on real cases, as so far reality has had an annoying habit of being elaborately odd in unexpected ways!

andylolz:
> I had thought that the Wikipedia scraper was mostly just for getting Party information?

That’s right, yes… But (as I understand it) party membership is for a particular term i.e. `on_behalf_of_id` is a property of a `Membership`.

If I try to do a `type: person`, the output data is really muddled. The `on_behalf_of_id`s don’t correspond to the input data at all.

tmtmtmtm:
Yeah — I think this is just a case we haven't had to deal with before, so isn't implemented yet. It's not something you should work around — we should expand the Merger to make it do the right thing. I'll have a look it tomorrow.

andylolz:
> then you should be using the Person merge, which is the part that's gradually being factored out into the CSV-Merge class further up

Ahh right, I seeee.

Sorry, so: Yes, I am indeed just adding (or merging) columns. No new rows are being added here. I _think_ the stuff above still applies – I need the option to merge based on multiple fields… Or (less generally) a merge that prioritises matching `term`s.

andylolz:
> we should expand the Merger to make it do the right thing. I'll have a look it tomorrow.

Great – thanks! I’ll leave this for now.

tmtmtmtm:
>Yes, I am indeed just adding (or merging) columns. No new rows are being added here. I think the stuff above still applies.

Yep — as you deduced above, so far when we've been adding columns they've been fields that apply to the Person, not to the Membership, so we need to extend it to be able to cope with that. It'll require a bit of thinking to work out the cleanest way to specify that in the `instructions`. This is all slightly trickier than might be expected as effectively we're trying to disguise a lot of complexity and shoehorn a lot of information into a format that isn't really suitable (i.e. making it so that someone could just create the information in Excel). In 90% of cases it works really well, but edge-cases can get a little awkward. IMO it's much better for us to find plausible ways to hide hide that, even if it's quite a lot more work, than to make the data creator jump through more elaborate hoops. Unfortunately you've gotten caught up in the middle of a case we haven't coped with yet.

Issue #604, opened 2015-08-28
Allow fallback patch strategies


tmtmtmtm:
e.g. OpenAustralia data is missing lots of wikipedia links. The ones that they _do_ have, we should do an exact match on, but the ones that don't, we should then be able to re-run with an approximate name match.

tmtmtmtm:
A simpler version of this could be that even just with name matching, try an exact match before an approximate one (e.g. in Belgium where all but 4 are exact matches between Wikipedia and the official site)

tmtmtmtm:
Duplicate of https://github.com/everypolitician/everypolitician/issues/142

PR #858, opened 2015-09-17
Gibraltar: initial data

### LINKED ISSUE #284: Gibraltar

tmtmtmtm:
Scrapers at https://morph.io/andylolz/gibraltar-parliament, https://morph.io/andylolz/gibraltar-wikipedia, and https://morph.io/tmtmtmtm/gibraltar-parliament-wikidata

Closes #284


tmtmtmtm:
@andylolz I _think_ this is sufficient. I cleaned up the old-style matching a little in #857, and then added a new `term_match` flag for add-a-column People matching. It looks like it's doing the right thing (e.g. E M Britto is a Member of Association for the Advancement of Civil Rights in 6, Gibraltar Social Democrats for 7-9, and 'unknown' in 10 & 11), but if you can give it a quick look-over before I merge that would be good.

If there's anything off, feel free to just cherry-pick anything you need from here back into your branch, tweak at will, and repush.

andylolz:
Looks good!

It doesn’t _quite_ get the result I expected, though I see why. The fuzzymatch happens first, then the term_match. I’m wondering if those operations should happen in the other order.

I’ve tweaked the gibraltar-parliament scraper so this gives the desired result either way, and have pushed that in #833.

tmtmtmtm:
> The fuzzymatch happens first, then the term_match. I’m wondering if those operations should happen in the other order.

That's probably a good idea. What did it get wrong this way around?

andylolz:
I’m checking “J J Bossano”, whose party should be showing as `gibraltar_socialist_labour_party` for terms 5–9 and 12. However, for some of those terms, he’s listed as “J Bossano”, which is of course not the best fuzzymatch score overall (since “J J Bossano” matches “J J Bossano” perfectly.) but it *is* the best fuzzymatch score if you limit by terms first.

tmtmtmtm:
Hmmm. Yeah, that's a slightly awkward one. As we have a suitable workaround I'm minded to leave this as-is until we hit another example of it going wrong. At the minute we pre-populate the 'haystack' to search against at the beginning of the process, as doing that for each match is hugely expensive. This way we'd need to pre-create a separate haystack for each Term, and then ensure we search against the correct one which would complicate the logic significantly more than this one-line version!

I do plan to switch the fuzzy-match searching to try an exact match first, and then only fall back to the approximate match where there are no results (#604), which should also potentially help with cases like this (and hopefully speed the process up considerably too!)

Are you happy with the currently staged version?

andylolz:
Yep, totally agree with all that!

Staged version looks good to me :rocket:

tmtmtmtm:
Yay! This one turned out to be _way_ more complicated than expected. Thanks for perservering!

PR #870, opened 2015-09-21
[Ukraine] Add faction history and faction IDs

henare:
This includes the faction history of deputies and also includes a unique ID for factions that comes from the source data.


tmtmtmtm:
Woohoo. Thanks @henare 

Would you mind also running the `bundle exec rake countries.json` step from the base directory (#3 in the [README](https://github.com/everypolitician/everypolitician-data/blob/master/README.md)? That should trigger a bunch of other things, including staging a display version of this.

(AFAIK github still doesn't let me add commits to someone else's PR, so I'd need to fork it to my own branch etc to do this myself)

henare:
>Would you mind also running the bundle exec rake countries.json step from the base directory (step 3 in the 'Building the data' section of the README? That should trigger a bunch of other things, including staging a display version of this.

Ahh, I hadn't seen that step - done!

You might want to hold of merging this until we fix https://github.com/openaustralia/ukraine_verkhovna_rada_deputies/issues/15

tmtmtmtm:
Thanks @henare. Yes — I can see some issues there (https://everypolitician-viewer-pr-688.herokuapp.com/ukraine/verkhovna-rada/term-table/8.html) e.g. with Березенко Сергій Іванович

henare:
Thanks for checking out the data @tmtmtmtm. I'll fix https://github.com/openaustralia/ukraine_verkhovna_rada_deputies/issues/15 and update this PR.

Basically it's the situation where someone joins parliament without a faction then goes and joins one. In the example you provide above they join the faction the day after joining the parliament.

henare:
@tmtmtmtm I've refreshed the data after fixing openaustralia/ukraine_verkhovna_rada_deputies#15.

Luckily your example above is interesting because according to the official data he [enters the house](http://itd.rada.gov.ua/mps/info/page/17568) on the 2nd of September and then ["joins" a faction](http://w1.c1.rada.gov.ua/pls/site2/p_deputat_fr_changes?d_id=17568) on the 3rd.

As you can see in my fix I'm setting the `end_date` to be the day before someone starts in the new faction. However I think this isn't displaying correctly in EP. Should I just set the `end_date` to the same date as the `start_date` of his new faction membership instead?

henare:
:point_up: ignore me - the data looks good. It just hadn't refreshed in the EP preview! (which is very fancy BTW).

tmtmtmtm:
Yep — looks good. Merged!

The preview has been a bit fiddly to get working right, particularly from external PRs, but it's super-helpful!

Issue #543, opened 2015-08-14
Italy: Historic Data


tmtmtmtm:
https://it.wikipedia.org/wiki/Deputati_della_XVII_Legislatura_della_Repubblica_Italiana & prev

briatte:
See also

http://dati.camera.it/ (XML/RDF)

and

http://storia.camera.it/

and for the upper chamber

http://dati.senato.it/

tmtmtmtm:
For the lower chamber, I can find scrapable lists per Term at http://storia.camera.it/deputati/faccette/leg_repubblica:16, http://storia.camera.it/deputati/faccette/leg_repubblica:15, etc, but I'm having difficulty getting a structured version. The linked RDF from each person there is useful, but doesn't seem to include the Parliamentary Group/Party info: e.g. http://dati.camera.it/ocd/persona.rdf/p300430

http://dati.camera.it/it/download/anagrafica.html looked plausible, but there were only 67 people in the download! 

I'm assuming it's possible to construct the perfect query from http://dati.camera.it/sparql, but my sparql knowledge is rather limited.

Best approach I can find so far is to start from http://dati.camera.it/ocd/gruppoParlamentare, and grab the memberships of each Parliamentary Group in turn. (Though it's not clear yet whether that would lose people who weren't a member of any Group, or whether there's a pseudo-Group for that)

Interestingly, though, those links give per-person data in a different format (e.g. http://dati.camera.it/ocd/deputato.rdf/d305904_17) that _does_ nest the relevant memberships, so presumably there's a way of iterating the entire set of 'deputado' records in turn.

@briatte: have you successfully navigated all this yet?

briatte:
Unfortunately not: I mention the "dati" and "storia" sources in my [README](https://github.com/briatte/parlamento), but my code uses the raw HTML pages for MPs in legislatures 15–17, because I was unable to use the SPARQL search engine properly :(

Italian groups are complicated to understand, and there's tons of changes constantly. MPs with no party affiliation are coded "Misto", although some "Misto" groups can also be coalitions of small parties (e.g. autonomists and splinters from larger formations).

The good part is that http://dati.camera.it/ocd/gruppoParlamentare seens to list "Misto" groups, so you should be safe.

PR #873, opened 2015-09-22
Add camera.it scraper for Italy (House) data

andylolz:
Scraper her: https://morph.io/andylolz/italy-camera

This scraper fetches `gender` and `birth_date` info from camera.it. ~~It could also fetch `start_date` and `end_date` info (but I haven’t done that yet).~~ It fetches `start_date` and `end_date` too.


data/Italy/House/sources/instructions.json
```diff
@@ -11,6 +11,23 @@
       "type": "membership"
     },
     {
+      "file": "morph/camera.csv",
+      "create": {
+        "type": "morph",
+        "scraper": "andylolz/italy-camera",
+        "query": "SELECT * FROM data"
+      },
+      "source": "https://morph.io/andylolz/italy-camera",
```
tmtmtmtm: Would you mind setting this to the 'end' source — i.e. the site that gets scraped, rather than scraper itself? This isn't really documented yet, or even _used_ yet, but I hope to do https://github.com/everypolitician/everypolitician/issues/127 pretty soon, which will actually display all these `source` fields, and it would be better if they were where the data came from, rather than the mechanism used to get them (if that makes sense?)

tmtmtmtm:
Oooh, excellent, thanks!

NB: that there's also structured data of this all available (https://github.com/everypolitician/everypolitician-data/issues/543) — I just haven't worked out yet how to construct the correct queries against that to be able to back-fill lots of historic data too.

tmtmtmtm:
Getting the `start_date` and `end_date` would indeed be very useful, as the existing data is only for current members, not those who have already left mid-term. Is this picking up everyone who has been a member during the term?

It would also be good if you could grab the initialism form of the Party name too and make it the `party_id`, and likewise for `area_id`s (if that's what the number before, e.g. XIV (MARCHE), means?)

We should ideally also expose the 'official' ID somewhere too, rather than losing it in the merge. Any column of the form `identifier__XXX` will be picked up, so you could either duplicate the ID across to a column of something like `identifier__parlamento_italiano` in Morph, or even just change the `query` in the instructions file to something like "SELECT *, id AS identifier__parlamento_italiano FROM data"

data/Italy/House/sources/instructions.json
andylolz: Oh yeah, sure. This was a copy/paste error!

andylolz:
> NB: that there's also structured data of this all available

Gah! You mentioned the structured data version to me before, but I completely forgot!

In that case it’s probably not _really_ worth continuing with this scraper… I’ve finish it off just as a temporary measure, but I’ll try to figure out the right SPARQL queries to use the structured data platform instead.

> Is this picking up everyone who has been a member during the term?

It is indeed! I’ve now added start and end dates in (I’m guessing parliament started on 5 March 2013 since that’s the most common start date, so I ignore that).

> if you could grab the initialism form of the Party name too and make it the party_id

I’m not sure where this is available…

> and likewise for `area_id`s

Ah, good point. I hadn’t spotted that. Yes, they do appear to be area IDs. Here’s the full list:

| ID | Name |
|---|------|
| A | EUROPA |
| B | AMERICA MERIDIONALE |
| D | AFRICA, ASIA, OCEANIA E ANTARTIDE |
| I | PIEMONTE 1 |
| II | PIEMONTE 2 |
| III | LOMBARDIA 1 |
| IV | LOMBARDIA 2 |
| V | LOMBARDIA 3 |
| VI | TRENTINO-ALTO ADIGE |
| VII | VENETO 1 |
| VIII | VENETO 2 |
| IX | FRIULI-VENEZIA GIULIA |
| X | LIGURIA |
| XI | EMILIA-ROMAGNA |
| XII | TOSCANA |
| XIII | UMBRIA |
| XIV | MARCHE |
| XV | LAZIO 1 |
| XVI | LAZIO 2 |
| XVII | ABRUZZO |
| XVIII | MOLISE |
| XIX | CAMPANIA 1 |
| XX | CAMPANIA 2 |
| XXI | PUGLIA |
| XXII | BASILICATA |
| XXIII | CALABRIA |
| XXIV | SICILIA 1 |
| XXV | SICILIA 2 |
| XXVI | SARDEGNA |
| XXVII | VALLE D'AOSTA |

So should I just store the area ID here? Or both bits?

tmtmtmtm:
> I’m guessing parliament started on 5 March 2013 since that’s the most common start date, so I ignore that

In general: If the `start_date` of the membership matches the `start_date` of the term, then it all gets sorted out in the merge anyway, so you don't need to do anything special with those (other than making sure that the `terms.csv` has the same date!)

>> if you could grab the initialism form of the Party name too and make it the party_id

> I’m not sure where this is available…

In any I've looked at it comes after the person's name on their page — e.g. on http://www.camera.it/leg17/29?shadow_deputato=306053&idpersona=306053&idlegislatura=17 where we have "MURA Romina - PD" (for PARTITO DEMOCRATICO), or http://www.camera.it/leg17/29?shadow_deputato=300306&idpersona=300306&idlegislatura=17 = "SANTELLI Jole - FI-PDL", etc.

> should I just store the area ID here? Or both bits?

Simplest answer is just provide the `area` and `area_id` for both, and it'll all get sorted out at merge.

More advanced answer is turn them into [OpenCivicData divisions](https://github.com/opencivicdata/ocd-division-ids) (see, for example what we do with Algeria, Australia, Haiti, Jamaica, Sweden etc) — this is what we're (slowly) moving to for all Area IDs.

andylolz:
Okay, I’ve made those updates. Since the new scraper fetches more members, I’ve made it the primary scraper in the instructions.json file. I hope that’s okay.

I haven’t converted the area IDs from roman numerals – should I do that?

tmtmtmtm:
>  Since the new scraper fetches more members, I’ve made it the primary scraper in the instructions.json file. I hope that’s okay

Great — I was going to suggest doing likewise. 

> I haven’t converted the area IDs from roman numerals – should I do that?

No — better to just leave them 'as is', I think. 


I'm finding it quite hard to keep track of these changes because the rebasing (in general it might be better only do that right at the end — or even not at all — other than for tiny fixups) — is this ready now? If so I think you still need to re-build the `countries.json`, so it'll trigger an update over on https://github.com/everypolitician/viewer-sinatra/pull/690

andylolz:
> I'm finding it quite hard to keep track of these changes because the rebasing

Ahh, apologies! I wasn’t exactly sure what the best thing was. I struggle to read a diff on a fixup for data (if you see what I mean) so I decided squashing would be easier.

> is this ready now?

I think so, yeah.

> If so I think you still need to re-build the countries.json, so it'll trigger an update

Aha – I hadn’t realised that’s what triggers the updates! Done.

tmtmtmtm:
>  I struggle to read a diff for a fixup of a data file (if you see what I mean) so I decided squashing would be easier.

Yeah — makes sense. Maybe keep the full history of changes to the config, and always just have the most recent data generated as a result?

> Aha – I hadn’t realised that’s what triggers the updates! Done.

Thanks. From the Viewer side, the *only* thing that matters is the `country.json` file, as is it's "map" for where to find data for each country. Everything is based off commit SHAs the whole way down. That's what makes the per-PR staging site so trivial: as long as there's a public commit _somewhere_, we can point at that SHA, not just when it's been merged to master.

So the [change](https://github.com/everypolitician/viewer-sinatra/commit/5f9bd266f0e9eee9b5f2f469de8db455849c7c20) opened on that repo is literally just changing the SHA of the `countries.json` you now want to use. And then the Sinatra app fetches the data for each country [based on the pointers within that file](https://github.com/everypolitician/viewer-sinatra/blob/66a917d838e7f69bf9f714e637ab7d87a19b9119/app.rb#L16-L17)

It's one of those things that's quite tricky to understand at first, largely because it's actually so simple. It seems like there should be something much more complex happening :)

andylolz:
\**nods and pretends to understand*\* haha okay I sort of get it.

> keep the full history of changes to the config, and always just have the most recent data generated as a result

Ah kk, good plan.

tmtmtmtm:
> *nods and pretends to understand* haha okay I sort of get it.

Hah. It's useful for me to try to explain it too, so we can find the best way to document it all :)


It might help to think of the app as being explicitly a way to view any EveryPolitician data at any point in time (compared to most apps that are only show _current_ data). The 'live' is site is simply one specific version of that, that happens to be set pointed at the more recent version of the data. And the staging sites are pointed at potential future versions of the data. But others could be viewing historic versions too. 

The app knows nothing about storage — when it wants the data for a country, it just wants a URL [from which it can load that](https://github.com/everypolitician/viewer-sinatra/blob/66a917d838e7f69bf9f714e637ab7d87a19b9119/lib/popolo_helper.rb#L8-L11). And the normal way of getting that is to have it referenced by SHA inside a `countries.json` file (which itself is simply loaded from the URL supplied in the `DATASOURCE` file). 

Want to see what the site looked like before the GLOW push? Simply find the `countries.json` from then (e.g. from PR #715) — and set the `DATASOURCE` for the app to be [that version of the file](https://github.com/everypolitician/everypolitician-data/blob/c81246006602750012f3b789a946b266c223541c/countries.json)

(BTW, this is how we built the animation for the [final blog post](https://www.mysociety.org/2015/09/15/everypolitician-200-countries-and-counting/) — I used `git log` to find the commits that added each country, and iterated through starting a local version of the app pointed at each of those SHAs in turn)

And if you really wanted, you could even do it pick-and-mix at a per legislature level to generate a site that never actually existed, but could have. Pick any SHA for each legislature, build a new `countries.json` file pointing at each of those, and then point the app at that new file.  Tada!

andylolz:
So I understood it as being similar to git submodules (in that just a SHA is stored). It differs, though, in that viewer-sinatra points to a particular version of a file that then contains lots more SHAs (one for each legislature.) But I suspect this is because it gives finer control, e.g. you can do the pick-and-mix thing you describe quite easily.  For instance, if you want to rollback one country’s data, I guess it’s as simple as changing one SHA in countries.json and then the bot does the rest.

tmtmtmtm:
Thanks @andylolz — this is now merged. I've also written a new pair of scrapers to get the Wikidata info too (https://morph.io/tmtmtmtm/italy-legislature-XVII-wikidata), but we can look at merging that separately.

tmtmtmtm:
@andylolz — I've had to unmerge this again, as the party info is quite messed up.

Some of this is my fault for not noticing earlier what you were doing with party info, but effectively you have the `party` as what they were elected as, but the `party_id` of their current status — thus completely at odds any time someone has switched (as many of them have!)

So, for http://www.camera.it/leg17/29?shadow_deputato=305606&idLegislatura=17, you have a party of "MOVIMENTO 5 STELLE BEPPEGRILLO.IT" and a party_id of "PD". But although they were _elected_ as M5S, they've since changed to PD, so these are mismatched.

Really we would want four separate membership entries here, similar to the example at http://everypolitician.org/submitting.html:

* MOVEMENT 5 STARS from March 19, 2013 to January 27, 2015
* Independent from January 27, 2015 to February 5, 2015
* ALTERNATIVA LIBERA from February 5, 2015 to April 30, 2015
* DEMOCRATIC PARTY  from April 30, 2015

(Some of this is also due to Italy having a stronger distinction between parties and parliamentary groups than you may be used to. Here I would distinguish the two carefully — use a separate column like `election_list` for what you're currently putting in `party`, and then use `group` or `faction` for the parliamentary group information.)

andylolz:
Ah, okay. Apologies for not spotting this :(

> use a separate column like `election_list` for what you're currently putting in `party`, and then use `group` or `faction` for the parliamentary group information.)

Okay, I’ve made that update; morph is re-running the scraper now.

I’m not exactly sure where the `group_id` should come from now. There isn’t a 1:1 mapping between the IDs shown next to the name and the group names (for instance, MISTO (mixed) is the ID given for various group names) and we don’t have an ID for many of the former groups. I can use the `idGruppo` param used in group page URLs ([for instance](http://www.camera.it/leg17/217?idlegislatura=17&idGruppo=1851)) but again, these pages aren’t always present for older groups.

tmtmtmtm:
> Ah, okay. Apologies for not spotting this :(

No worries — it's easy to miss. I only saw it when I noticed that a party had entirely vanished from the seat count summary at the top of the term page.

> I’m not exactly sure where the group_id should come from now

If there's nothing obvious, it's OK to just omit it, and we'll work out later how to add them for the ones for whom it makes sense. I'd like to do a big push generally on getting deeper Party/Group info in the not too distant future. (https://github.com/everypolitician/everypolitician/issues/145)

andylolz:
I’ve tried rebuilding the data… One thing I notice is “LEGA NORD E AUTONOMIE” and “Lega Nord e autonomie” are listed as different organizations, although the auto-generated IDs are identical, so it sort of works out.

tmtmtmtm:
Yeah, the code to auto-generate IDs is deliberately quite simplistic. If there's a case where it doesn't work, then that should be resolved upstream to EP itself (e.g. in individual scrapers) where it's easier to work out the correct thing to do.

PR #890, opened 2015-09-24
Italy (House): second attempt at camera.it scraper

andylolz:
Gah, this looks better but I’m still not sure the groups are quite right.


tmtmtmtm:
@andylolz do you have specific concerns? The breakdowns looks pretty similar to the [currently live version](http://everypolitician.org/italy/house/term-table/17.html), other than the group names being somewhat longer, and the 'Misto' ones being subdivided further (which seems like a bonus to me!)

The main oddities I can see are:
* http://www.camera.it/leg17/29?shadow_deputato=302741&idLegislatura=17 
* No-one (other than this case) starting until 2013-03-19, which makes me think we should probably set that as the start_date of the Term

andylolz:
> No-one (other than this case) starting until 2013-03-19, which makes me think we should probably set that as the start_date of the Term

Ahh – I’d meant to add a “no party” period for everyone (between their “Proclamato” date, and the date they first join a group.) But perhaps that isn’t interesting, really.

tmtmtmtm:
I think that over time we'll probably need to evolve something whereby Terms can have two separate start dates, but for now I'd be tempted to just set the start date to the date when all the Groups get recognised, as presumably there can't really be much in the way official business before then anyway. 

This is one of these areas where I think it's more important to have consistency as the primary feature, rather than strict accuracy — people shouldn't need to become an expert on the system of each country just to be able to use the data, or to have to have lots of `if country == X … else if country == Y` throughout code that operates across multiple jurisdictions. We can simply declare what _our_ `start_date` for a term means, and have that mean the same thing everywhere, and then have a separate field on a country by country basis that people with local knowledge can use instead where that distinction is important to them.

andylolz:
All makes sense, yep. I’ve made that change now. [Looks good!](https://everypolitician-viewer-pr-701.herokuapp.com/italy/house/term-table/17.html)

The thing I wasn’t really happy about was that there was a group (Lega Nord e autonomie) being declared twice in the json. It was actually a result of the merge, but in any case it seems to have resolved itself now.

One other bit of potential weirdness is that now [Matteo Salvini](http://www.camera.it/leg17/29?shadow_deputato=302741&idLegislatura=17) has an `end_date` that predates the start of the term. Leave that, or change it to 2013-03-19?

Incidentally, these group memberships change remarkably often! There appear to have been 17 new changes since I started working on this bot?! I re-ran it because I didn’t think that could possibly be correct!

tmtmtmtm:
Yay!

> One other bit of potential weirdness is that now Matteo Salvini has an end_date that predates the start of the term. Leave that, or change it to 2013-03-19?

I think it's OK to leave it. I'm assuming this a case of someone being elected but never actually taking their seat (staying as an MEP instead)? 

> Incidentally, these group memberships change remarkably often! 

Yep! Some of the European countries have a LOT of switching going on.

You happy for this to be merged now? Are you still hoping to do something with SPARQL to get old data too?

andylolz:
> You happy for this to be merged now?

I think so, yes. I would like to fix the group IDs at some point, but it’s not really obvious how to do that at the moment.

> Are you still hoping to do something with SPARQL to get old data too?

Yes, I shall take a look at that! It might possibly make all of this work redundant… But I understand the Italian system a bit better now, anyway!

tmtmtmtm:
Yay. Merged.

> I would like to fix the group IDs at some point, but it’s not really obvious how to do that at the moment.

The RDF gives a 'dcterms:alternative' for them: e.g. http://dati.camera.it/ocd/gruppoParlamentare.rdf/gr1611

So it should all shake out OK if you manage to get that working.

andylolz:
> The RDF gives a 'dcterms:alternative' for them

Ah, excellent!

Issue #986, opened 2015-10-09
Guinea Bissau: Missing one person


chrismytton:
One of the members is missing because there appear to be two people called Daniel Gomes in the same party and we aren't generating ids for each person. Related to #981

PR #987, opened 2015-10-09
Guinea Bissau: refresh data

chrismytton:
Refresh after adding `id` field to scraper.

Fixes https://github.com/everypolitician/everypolitician-data/issues/986


tmtmtmtm:
This also reveals a bug with names.csv, in that it only shows distinct names, rather than distinct name-id pairs…

Issue #690, opened 2015-09-03
Trinidad and Tobago


tmtmtmtm:
http://ttparliament.org/

Maddock-:
All previous parliaments easy to parse here: http://ttparliament.org/members.php?mid=26

An election took place yesterday, the results are available in a lovely parsable pdf... 
http://www.ebctt.com/wp-content/uploads/Election-Agents-20050906.pdf 

Or on the only slightly better Wikipedia article:
https://en.wikipedia.org/wiki/Trinidad_and_Tobago_general_election,_2015

I imagine it might be more productive to wait for the government website to be updated and use the same parser!

(!) the results are on Wikipedia

tmtmtmtm:
Yep, I'll leave this a while and see if the official data updates. Thanks.

tmtmtmtm:
Closed by PRs linked above.

Issue #102, opened 2015-05-20
Trinidad and Tobago


duncanparkes:
There is an election coming up for the House of Representatives. Try to capture the old data before it's too late (if it's not already).

duncanparkes:
I don't think we need to worry about losing the data - they have archived data on the web going back to the 50s! I suggest leaving this until after the election and concentrating on places with upcoming elections where they only show the present representatives on the site.

duncanparkes:
It's a great candidate to get some historic data as soon as we run out of those though.

PR #1096, opened 2015-10-26
Trinidad and Tobago: Representatives

chrismytton:
Scraper here https://morph.io/jennahowe/trinidad_and_tobago

Part of #102 and #690


tmtmtmtm:
This is a slightly odd one. The term ID in the incoming data doesn't match the one in the terms file, so it doesn't match up well (and thus doesn't display the term dates). But there's some code somewhere that tries to guess what the term _should_ be if there is none, so it gets a half-hearted version. We should probably just make it abort in those cases, so that things like this don't get through, but for now you should either change the ID in the `terms.csv`, or (perhaps better) override the incoming term by changing the SQL (`SELECT *, 11 AS term FROM data`)

chrismytton:
@tmtmtmtm Ah good point. I've updated the SQL as you suggested and it's picked up the dates correctly now. I've also made the same change in https://github.com/everypolitician/everypolitician-data/pull/1097 for the Senate.

chrismytton:
Thanks @jennahowe for writing the scraper for this! :tada:

PR #1125, opened 2015-10-29
Reconciliation with html interface

chrismytton:
This changes the reconciliation process over to using an html interface. It also adds a file which maps a source's ids to UUIDs.

The dragging and dropping doesn't work properly yet, but it's got enough functionality to replace the existing CSV based interface.

Part of https://github.com/everypolitician/everypolitician/issues/195


tmtmtmtm:
It might be the speed of my internet connection (grrr Berlin!), but this doesn't work for me at all (unless there's something that I'm meant to do that isn't obvious to me):

screen shot 2015-10-29 at 20 45 44

templates/reconciliation.html.erb
```diff
@@ -0,0 +1,53 @@
...
+<%= reconciliation_js %>
+    </script>
+
+  </head>
+  <body>
+    <div class="container">
+      <div class="row">
+        <div class="col-md-12">
+          <h1 class="page-title">Reconciliation</h1>
+          <p>Check that the names on the left match the names on the right. If they don't then press the <code>x</code> to the right of the row and it will drop to the bottom. If the row matches someone in the existing data you can choose their name from the dropdown one the row appears at the bottom of the table.</p>
```
tmtmtmtm: Presumably this is meant to be "once the row"?

templates/reconciliation.html.erb
```diff
@@ -0,0 +1,53 @@
...
+          <table class="table">
+            <tr>
+              <th>Existing</th>
+              <th>Incoming</th>
+            </tr>
+          </table>
+          <select class="js-merged-rows" style="display: none;">
+            <option value="">Select an existing person to match with</option>
+            <% merged_rows.sort_by { |r| r[:name] }.each do |row| %>
```
tmtmtmtm: Shouldn't this do a uniq-by-UUID filter first (or is that already happening somewhere else that I can't see?)

templates/reconciliation.html.erb
chrismytton: Oops, yes, fixed.

templates/reconciliation.html.erb
chrismytton: Ah, yes, still keep forgetting these are memberships not people. Fixed in https://github.com/everypolitician/everypolitician-data/commit/ccf2d4f34b86fec28e669910179b3dfa7676d5ff.

chrismytton:
> It might be the speed of my internet connection (grrr Berlin!), but this doesn't work for me at all (unless there's something that I'm meant to do that isn't obvious to me):

Which country were you trying? I've tried a few random ones and they all seems to be working. I am loading quite a few external assets so that might indeed be a problem over a slow connection if they timeout/are blocked/whatever.

tmtmtmtm:
I tried several, and had the same result for all. Replacing jQuery with the minimised version from Google (https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js) helped a bit, but I didn't get as far as finding similar alternatives for the others.

chrismytton:
@tmtmtmtm I've switch to using Google's hosted version for jquery and jquery UI in https://github.com/everypolitician/everypolitician-data/commit/523345c25eba463a892fc4f9aa81765b79a8135f so hopefully that will help. I'm a bit reluctant to pull the javascript and css into this repo as it feels like unnecessary bloat. But that might be the best thing to do to ensure it works with slow/no internet.

templates/reconciliation.js
```diff
@@ -0,0 +1,84 @@
...
+        if (uuid) {
+          existing = $('select option:selected', $existing).text();
+        } else {
+          existing = '';
+        }
+      } else {
+        uuid = $existing.data('uuid');
+        existing = $existing.data('text');
+      }
+      csv.push([id, incoming, uuid, existing, null].join(','));
```
tmtmtmtm: You should quote the incoming and existing fields here in case they have commas in them. At the minute this will generate a line like:

```Q7372562,Roy Beggs, Jr.,9091a64b-281c-425a-b044-8691dcf18053,Roy Beggs,```

which isn't correct (the UUID is now set to " Jr.")

tmtmtmtm:
This is working now for me, and other than the bug with needing to quote/escape the name fields in the CSV, it's looking good!

templates/reconciliation.js
chrismytton: Ooh, good catch! I did wonder about using (the slightly bizarrely named) [Papa Parse](http://papaparse.com) to handle CSV oddities, but it seems like a bit of an overkill. I guess just manually quoting them should be ok, unless there are any fields that contain `"` already?!

tmtmtmtm: There doesn't seem to be any at the minute, based on a quick scan, and hopefully by the time any appear we'll only be outputting IDs and UUIDs here anyway…

PapaParse does look quite nice though…

chrismytton: Went with Papa Parse in the end because the code was *much* simpler! Fixed in https://github.com/everypolitician/everypolitician-data/commit/01dbdcafe026a6e5f06d2e788e54e5941c7ba15b

tmtmtmtm:
Looks good. I've also added the `reconciliation/*.html` files to `.gitignore` — no need to have those lying around once the reconciliation is done.

PR #1130, opened 2015-10-30
Netherlands: Wikidata

### LINKED ISSUE #778: Netherlands: Wikidata
via https://nl.wikipedia.org/wiki/Samenstelling_Tweede_Kamer_2012-heden

chrismytton:
Scraper here https://morph.io/chrismytton/netherlands-tweede-kamer-wikidata

Fixes #778


tmtmtmtm:
There's something slightly off with this, as the missing five people are quite easy to find (although "Tony P C van Dijck" is a little non-obvious as "Teun van Dijck")

With a bit of investigation, it seems the scraper is giving up too early — the last person it finds on the page is "Roelof Bisschop" — so I suspect it's something to do with single-column vs double-column entries.

chrismytton:
> With a bit of investigation, it seems the scraper is giving up too early — the last person it finds on the page is "Roelof Bisschop" — so I suspect it's something to do with single-column vs double-column entries.

Gah, yes, it looks like the `<ul>` elements after Roelof aren't wrapped in `<div class="editmode">` elements, so they're getting missed out.

PR #1165, opened 2015-11-04
Reconciliation ui

chrismytton:
HTML-based one-at-a-time reconciliation of incoming people to existing ones:
screen shot 2015-11-04 at 17 41 53

For now we only reconcile people for whom we can't find an exact match — later we'll turn it on for all incoming data.

Closes https://github.com/everypolitician/everypolitician/issues/195


tmtmtmtm:
Woohoo :rocket:

PR #1170, opened 2015-11-05
Switch to UUIDs

tmtmtmtm:
For all countries where the migration is clean (i.e. where we weren't previously generating our own IDs), switch to UUIDs, including the previous ID in the `identifiers`.

Part of https://github.com/everypolitician/everypolitician/issues/199


henare:
This just tripped up our importer so I came to find out why this change was made but I can't see anything. Is the reasoning outlined somewhere? :four_leaf_clover:

tmtmtmtm:
@henare — super sorry about that! We screwed up the implementation of this quite badly. 

Some of the reasoning is explained at https://github.com/everypolitician/everypolitician/issues/180, but we clearly should have introduced that in a much cleaner way. We're currently discussing how best to signal data changes in future. (We have a few ideas, but suggestions for what would work best for you are very welcome!)

The "old" ID is still available in the `identifiers` section of any Person in the `everypolitician_legacy` section, if you need to use that. If that ID is meaningful in some other way (e.g. it's an official ID from the Rada), then it can also be exposed under a different name by exposing it in the incoming CSV as an `identifier__[name]` column (which can also be simulated by altering the `query` in the instructions file: e.g. "SELECT *, id AS identifier__rada FROM data" — I'll make sure this is explained more fully somewhere in the next day or two, as the docs are unfortunately lagging behind at the minute.

Again, apologies for such a major change being unannounced — we'll try to ensure nothing like this happens again!

henare:
>@henare — super sorry about that! We screwed up the implementation of this quite badly. 

It's OK. Thanks for getting back to me.

We still haven't solved it downstream for our application (https://github.com/OPORA/publicwhip/issues/40).

>We're currently discussing how best to signal data changes in future. (We have a few ideas, but suggestions for what would work best for you are very welcome!)

I just looked around the repo at issues and pull requests. I also checked for a CHANGES or CHANGELOG file.

>If that ID is meaningful in some other way (e.g. it's an official ID from the Rada), then it can also be exposed under a different name by exposing it in the incoming CSV as an identifier__[name] column (which can also be simulated by altering the query in the instructions file: e.g. "SELECT *, id AS identifier__rada FROM data" — I'll make sure this is explained more fully somewhere in the next day or two, as the docs are unfortunately lagging behind at the minute.

It does seem to be the official ID used by the Rada. I've opened #1230 to track this change.

>Again, apologies for such a major change being unannounced — we'll try to ensure nothing like this happens again!

The only thing constant is change :)

PR #1173, opened 2015-11-05
Generate UUIDs for legislatures with no IDs

tmtmtmtm:
If an incoming source doesn't have an ID column, generate a pseudo-ID instead. This allows us to correctly merge later IDs, so as to expose the correct legacy ID, whilst also generating and exposing UUIDs.


rake_helpers/combine_sources.rb
```diff
@@ -211,10 +211,13 @@ def combine_sources
         id_map = Hash[CSV.table(ids_file, converters: nil).map { |r| [r[:id], r[:uuid]] }]
       end
       table = csv_table(file)
-      abort "No id column found for #{src[:file]}" unless table.first.keys.include?(:id)
+
       table.each do |row|
+        # If the row has no ID column, we'll need something we can treate as one. 
+        # This 'pseudo id' can be specified explicitly, or defaults to 'name'
+        row[:pseudoid] = row[:name].downcase.gsub(' ','_') unless (row[:id] || row[:pseudoid])
```
chrismytton: Probably best to make this identical to [how csv_to_popolo worked](https://github.com/tmtmtmtm/csv_to_popolo/blob/ba0af5638e94438b6b5ffc5b4364ba135aa645d1/lib/csv_to_popolo.rb#L162).

chrismytton:
Looks good :+1:

PR #1185, opened 2015-11-06
Persist CSV data between reconciliation runs

chrismytton:
When there are new people to be reconciled persist the data so that only
the new people have to be reconciled rather than having to do everyone
again.

Closes https://github.com/everypolitician/everypolitician/issues/197

### DIFF
--- a/rake_helpers/combine_sources.rb
+++ b/rake_helpers/combine_sources.rb
@@ -261,7 +261,7 @@ namespace :merge_sources do
           end
           reconciler = Reconciler.new(merged_rows, merger, reconciled)
           need_reconciling = incoming_data.find_all do |d|
-            reconciler.find_all(d).empty? && !reconciled.any? { |r| r[:id].to_s == d[:id] }
+            reconciler.find_all(d).to_a.empty? && !reconciled.any? { |r| r[:id].to_s == d[:id] }
           end
           fuzzer = Fuzzer.new(merged_rows, need_reconciling, merger)
           matched = fuzzer.find_all.sort_by { |m| m.last }.reverse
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
This doesn't really work very well as it now requires you to have actually reconciled everyone in a legislature before it lets you continue with the build — you can't skip people until later…

chrismytton:
> This doesn't really work very well as it now requires you to have actually reconciled everyone in a legislature before it lets you continue with the build — you can't skip people until later…

Hmm yes good point. How do we want this to work then? I'm not sure we can currently differentiate between rows that you've skipped and new rows that have yet to be reconciled. I guess we could keep a separate record of ones you've skipped and include them in subsequent reconciliations when there are new people to reconcile?

chrismytton:
Have updated this so it only refuses to build if the reconciliation CSV doesn't exist. This means that the reconciliation html *always* gets generated so you can jump into the reconciliation at any time. I've also added some output telling you how many rows haven't been reconciled, which should make it easier to spot when new people have been added.

tmtmtmtm:
The flow of this now works well — however it falls over in, for example, Slovenia, with

```
rake aborted!
NoMethodError: undefined method `empty?' for nil:NilClass
everypolitician-data/rake_helpers/combine_sources.rb:264:in `block (3 levels) in combine_sources'
```

It could probably do with splitting some things out into well-named helper functions to make the flow a little clearer, as this is all getting quite debse (though I'm OK with doing some of that as gradual improvement too.)

I've also opened https://github.com/everypolitician/everypolitician/issues/203 to deal with the case where there's now nothing to reconcile and the HTML is a little bare — no real need to handle that here yet.

tmtmtmtm:
There's a bug with the progress bar seemingly counting based on the wrong information, but I've ticketed that separately (https://github.com/everypolitician/everypolitician/issues/205), and will merge this for now as it's still more useful than the current version.

PR #1225, opened 2015-11-11
Add script to run command for each legislature

chrismytton:
A handy script I've been using to check that rebuilding each legislature doesn't error after I've made a change to the rake scripts.


tmtmtmtm:
:+1:

PR #1238, opened 2015-11-12
Convert legacy reconciliation CSVs in place

chrismytton:
Convert legacy reconciliation CSVs that contain an id.

Converting reconciliation files that have something like `incoming_name,existing_name,confidence` in the header is going to be harder because there is no `id` in the incoming data, so there's no obvious way to map it in the new style reconciliation files, which have `id,uuid` headers.

Fixes https://github.com/everypolitician/everypolitician/issues/209

### DIFF
--- a/rake_helpers/combine_sources.rb
+++ b/rake_helpers/combine_sources.rb
@@ -68,7 +68,7 @@ class Reconciler
     @_lookup_by_uuid ||= @_existing_rows.group_by { |r| r[:uuid].to_s }
   end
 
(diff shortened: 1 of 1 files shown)


chrismytton:
When a legacy reconciliation CSV is encountered it will automatically convert it in-place. Once all the CSVs have been converted this code can be removed again.

tmtmtmtm:
> Once all the CSVs have been converted this code can be removed again.

Would be good to link here to the ticket that's been opened to remind us to remove this again…

PR #1246, opened 2015-11-12
Belgium: switch to new-style reconciliation file

tmtmtmtm:
Switch to ID/UUID and refresh data


tmtmtmtm:
There's something very odd going on with this one — needs further investigation, probably together with @chrismytton

chrismytton:
Have updated this with the new-style reconciliation file, generated by the code added in #1238 and #1443,

tmtmtmtm:
Yay!

PR #1249, opened 2015-11-12
Croatia: switch to new-style reconciliation file

tmtmtmtm:
Switch to ID/UUID, and refresh data

DO NOT MERGE: this is another odd one, where most of the records that appear in the prior reconciliation file aren't being offered when reconciling. Need to discuss with @chrismytton.


chrismytton:
Updated to new-style file with code added in #1238 and #1443

PR #1274, opened 2015-11-13
Get more group information from Wikidata

chrismytton:
This uses [wikisnakker](https://github.com/everypolitician/wikisnakker) to bulk fetch information about parties from Wikidata and adds it to the resulting popolo. I've also added some tests for this, which having laid the foundation will make it easier to test other bits and bobs in the future, e.g. the reconciliation interface.

- [x] Update to use Wikisnakker master once the use of `recurse_proc` has been tidied up.
- [x] Refer to groups instead of parties
- [x] Expose the Wikidata ID in the resulting Popolo

Part of https://github.com/everypolitician/everypolitician/issues/145
Closes https://github.com/everypolitician/everypolitician/issues/206


tmtmtmtm:
This looks great, though I suspect we should use the terminology of 'group' everywhere rather than 'party', for consistency with Popolo (and because in many countries these will be Factions etc rather than Parties). 

Probably important to also expose the Wikidata ID in the resulting Popolo right from the start too.

There's also a slight inconsistency in using `wikidata_id` as the field name here, when we just use plain `wikidata` when doing likewise for the Legislature, or Term. (This may be a sign we should change those, rather than this, but being the same everywhere is generally a Good Thing).

tmtmtmtm:
Woot :+1:

PR #1301, opened 2015-11-14
Sweden: Merge Wikidata party info

tmtmtmtm:
Add party information from Wikidata

@chrismytton — this one also blows up. I dug into it more deeply, and
discovered it’s because Q10177454 doesn’t have an English label. This
causes 'populate_with' to blow up, as it assumes that everything will.


chrismytton:
@tmtmtmtm This is now also fixed, updated and ready to go!

PR #1446, opened 2015-11-19
Sri Lanka: switch to new-style reconciliation file

chrismytton:
Switch to ID/UUID


tmtmtmtm:
This one happened naturally as part of #1464

PR #1452, opened 2015-11-19
Indonesia: switch to new-style reconciliation file

chrismytton:
Switch to ID/UUID


tmtmtmtm:
This one seems to be a duplicate of #1444

PR #1474, opened 2015-11-20
Disable converters for reconciled CSVs

chrismytton:
We want to treat all cells in the reconciliation CSV as strings, so we
disable all converters to avoid any surprises.

This should fix the problem @tmtmtmtm was having in https://github.com/everypolitician/everypolitician-data/commit/31fefa1d3fccfee803da3b3a0876282fc8b6d2e4.


tmtmtmtm:
Yep — we really need to find a way to prevent this sort of thing in future. This at least the 3rd time we've been caught out by this one…

Issue #1215, opened 2015-11-10
Tidy up the reconciliation code


chrismytton:
At the moment [the reconciliation code](https://github.com/everypolitician/everypolitician-data/blob/96de08c33f0b4ad42eca4d0b94aad6af0b34342e/rake_helpers/combine_sources.rb#L250-L276) is a bit hard to follow.

https://github.com/everypolitician/everypolitician-data/pull/1185 changed things so that the CSV is persisted between runs but it also slows things down because it checks the reconciler for every row, which in places where we have a lot of data, e.g. Turkey, this can slow things down.

It would be good to refactor this to make it a bit easier to understand and change.

PR #1487, opened 2015-11-20
Reconciliation improvements

chrismytton:
Moves most of the reconciliation logic out of the `rake_helpers/combine_sources.rb` file and into separate classes in the `lib` directory. This makes it easier to focus in on particular bits of the code for refactoring, rather than having to comprehend what the whole blob of code is doing.

I've also added some extra logic to the reconciliation interface which displays fields which are common to both the existing and incoming fields, including images.

screen shot 2015-11-20 at 17 43 16

Related to https://github.com/everypolitician/everypolitician/issues/204

Closes https://github.com/everypolitician/everypolitician-data/issues/1215

### DIFF
--- a/lib/reconciliation.rb
+++ b/lib/reconciliation.rb
@@ -1,3 +1,4 @@
 require_relative './reconciliation/interface'
 require_relative './reconciliation/fuzzer'
 require_relative './reconciliation/legacy_csv'
+require_relative './reconciliation/template'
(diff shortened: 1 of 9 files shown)


chrismytton:
Just to note I've tested running `bundle exec rake clean default` on all legislatures using the `bin/run_command_for_each_legislature` script. It ran cleanly, the only changes were in ordering of items in the CSV for a couple of countries, but the actual data stayed the same.

tmtmtmtm:
This looks great. Interestingly, it's already exposed a quite severe bug, but wasn't obvious before! (#1491). But that already existed, and this will probably make it easier to track down, as well as making all other merging a lot simpler too.

PR #1566, opened 2015-11-24
Combine AreaWikidata and GroupWikidata classes

chrismytton:
These classes were basically identical, so for now I've combined them.
In the future there may be some differences, and when there are we can
create subclasses or change things around to use composition. But for
now this is better than having duplicated code.

For reference, this was the difference between the classes before I combined them:

```diff
--- lib/group_wikidata.rb       2015-11-19 12:16:12.000000000 +0000
+++ lib/area_wikidata.rb        2015-11-24 09:13:17.000000000 +0000
@@ -1,8 +1,10 @@
 require 'wikisnakker'
 
-# Takes an array of hashes containing group 'id' and 'wikidata' then returns
-# wikidata information about each group.
-class GroupWikidata
+# Takes an array of hashes containing area 'id' and 'wikidata' then returns
+# wikidata information about each area.
+# TODO: combine this with GroupWikidata
+
+class AreaWikidata
   attr_reader :wikidata_id_lookup
 
   def initialize(mapping)
@@ -12,10 +14,10 @@
   end
 
   def to_hash
-    group_information = wikidata_results.map do |result|
+    area_information = wikidata_results.map do |result|
       [wikidata_id_lookup[result.id], fields_for(result)]
     end
-    Hash[group_information]
+    Hash[area_information]
   end
 
   def wikidata_results
```


tmtmtmtm:
:+1: :rocket:

PR #1772, opened 2015-11-30
Fix link in readme

tijmenb:
The link to gender-balance.org is missing `http`, which means it clicking on it will lead the user to [a 404 page on GitHub](https://github.com/everypolitician/everypolitician-data/blob/master/gender-balance.org), and not the site.


tmtmtmtm:
Thanks @tijmenb!

PR #1780, opened 2015-12-01
Combine "Skip" and "No matches" buttons

zarino:
The new unified "No matches" button does what the old "Skip" button used to do (ie: adds a null row to window.votes, which is skipped by the CSV compiler).

Fixes everypolitician/everypolitician#216.

screen shot 2015-12-01 at 12 19 52

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -44,8 +44,7 @@
                 <div class="column-6 pairing__choices">
                     <h2>Potential matches from existing data</h2>
                     {{ existingPersonHTML }}
-                    <div class="no-matches">None of these match</div>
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
Would be good to keep the right-arrow-to-move-forwards behaviour, simply making it identical to hitting 0.

zarino:
Would you just prefer right-arrow instead of 0?

tmtmtmtm:
If we should only have one, then I'd definitely prefer the arrow to zero.

zarino:
@tmtmtmtm – done!

tmtmtmtm:
:+1: :rocket:

PR #1781, opened 2015-12-01
Ignore keyboard shortcuts when CSV tray is open

zarino:
Especially handy for using the right-arrow key to navigate the text inside the CSV textarea.

Fixes everypolitician/everypolitician#214.


tmtmtmtm:
:+1:

PR #1782, opened 2015-12-01
Show helpful message when there’s nothing to reconcile

zarino:
Fixes everypolitician/everypolitician#203.

screen shot 2015-12-01 at 12 22 59


tmtmtmtm:
:+1:

PR #1784, opened 2015-12-01
Colour-code existing matches based on match strength

zarino:
We use the Dice coefficient for each match to highlight strong matches, and de-emphasise weak matches. Middling matches (0.6-0.9) are pulled out in yellow.

screen shot 2015-12-01 at 16 22 20

Fixes everypolitician/everypolitician#201.

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -62,11 +62,15 @@
       {% } %}
           {% if (person.matchStrength) { %}
             <header class="person__meta">
-              <div class="person__match-strength">{{ person.matchStrength }}% match</div>
+              <h1>
+                {{ person[field] || person.name }}
+                <span class="person__match-strength">{{ person.matchStrength }}% match</span>
+              </h1>
               <div class="progress-bar"><div style="width: {{ person.matchStrength }}%"></div></div>
             </header>
-          {% } %}
+          {% } else { %}
             <h1>{{ person[field] || person.name }}</h1>
+          {% } %}
             <dl>
               {% fields.forEach(function(field) { %}
                 {% if(field !== 'id' && person[field]){ %}
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -54,7 +54,7 @@
 
     <script type="text/html" id="person">
       {% if(person.uuid){ %}
-        <div class="person" data-uuid="{{ person.uuid }}">
+        <div class="person {{ person.matchStrength }}" data-uuid="{{ person.uuid }}">
       {% } else if(person.id){ %}
         <div class="person" data-id="{{ person.id }}">
       {% } %}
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
I'm not convinced that this is majorly valuable like this as it doesn't show relative strengths, and is a little arbitrary on the cut-off-points. In the scenario where we're merging people with Spanish names, where one source has two names, and the other has four names, for example, a very strong match still might only score 0.5.

zarino:
Ok, now looks like this:

screen shot 2015-12-02 at 12 53 06

zarino:
There's scope for potentially changing the bar colour from green to orange or red based on how close that person's strength is to the other people’s strengths.

ie: the width of the bar shows the absolute match strength, but the colour shows the relative strength.

That way, a 30% bar followed by another 30% bar would both be presented in green (because they're the #1 and #2 matches, therefore still the best hits despite the absolute weakness). Whereas a 100% bar followed by a 30% bar would put the 100% bar in green and the 30% bar in orange or whatever, because, compared to the 100% match, that 30% match is pretty unlikely to be right.

tmtmtmtm:
oooh, that's very nice! The colour coding idea is nice too, but this version looks like it'll be good enough for now, and I'd rather have https://github.com/everypolitician/everypolitician/issues/234 than extra tweaks on this, I think. 

Only slight comment is that it's probably better to be explicit that the "90% match" is on the name, not the whole record.

zarino:
@tmtmtmtm done!

screen shot 2015-12-02 at 13 47 00

tmtmtmtm:
That's a much nicer solution that I was imagining! I thought it would just be textual "100% NAME MATCH" or something. But this version generally looks nicer too :)

zarino:
:+1:

andylolz:
This reconciler is looking __so cool__. Is this a public thing yet? Like, are there instructions somewhere on how to use it?

tmtmtmtm:
@andylolz — there isn't a public interface to it (at least not yet) — it's mostly an internal tool for when we're trying to merge data from multiple sources (e.g. mapping Wikidata to the official site). To test it out try going into `Israel/Knesset', run `bundle exec rake clobber default`, and then open `sources/reconciliation/wikidata.html` in your browser. Then you can try to work out whether anyone on the right hand side is a match for the person on the left…

(For the full experience you can also try `rm sources/reconciliation/wikidata.csv` before running the `rake` task — if you run against master it'll start from as far as we've got so far, whereas if you delete that you'll have to reconcile everyone one by one…)

Once finished, paste the CSV back into `sources/reconciliation/wikidata.csv` and open a Pull Request :)

andylolz:
Aha, gotcha! Thanks, @tmtmtmtm :)

[Btw I do watch this repo (via github notications) but it’s difficult as a casual observer to pick signal from noise. There are a few gems in amongst loads of (data-related) PRs. I’m just noting this here because it’s a pretty weak case for splitting data and code into separate repos! But I wonder if there are any plans to do that?]

PR #1804, opened 2015-12-02
Highlight suggestions that have already been matched

zarino:
Fixes everypolitician/everypolitician#215.

If you come across a suggested person who's already been matched to an incoming person (either by you in this sitting, or previously) then you get a little warning:
 
screen shot 2015-12-02 at 15 58 19

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -88,5 +88,11 @@
         </div>
     </script>
 
+    <script type="text/html" id="personAlreadyMatched">
+      <div class="person__already-matched">
+        Already matched to incoming “<b>{{ person[field] || person.name }}</b>”
+      </div>
+    </script>
+
 </body>
 </html>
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
Yay! This is excellent. Unfortunately it won't pick up on anyone who was matched to someone else by being an exact match, and thus not even part of the reconciliation process, but that seems like something that should be fixed elsewhere…

zarino:
@tmtmtmtm, oh good point – maybe we need a list of those people in something like `window.automaticMatches` or something?

tmtmtmtm:
I think it would be good to have those passed into the reconciler, and then written out to the CSV (so they can be hand-edited if they were wrong), but we'd want to skip them in the normal reconciliation process.

tmtmtmtm:
I've opened 235 to hand the wider case, so happy enough to merge this version for now, as it's a very useful addition even before that!

PR #1807, opened 2015-12-02
Remove the legacy_csv converter

tmtmtmtm:
We've now completed the migrations from https://github.com/everypolitician/everypolitician/issues/225, so we can remove this temporary scaffolding.

Closes https://github.com/everypolitician/everypolitician/issues/225


tmtmtmtm:
@chrismytton is this enough, or are there any other remnants lurking around that can removed with this too?

chrismytton:
@tmtmtmtm I think this is all that needs to be done, so this can be merged. :+1:

PR #1820, opened 2015-12-03
Chile: Merge Wikidata

tmtmtmtm:
Scraper at https://morph.io/tmtmtmtm/chile-congress-wikidata

Closes https://github.com/everypolitician/everypolitician-data/issues/1056


tmtmtmtm:
@zarino there's a nasty bug somewhere here. I reconciled the first 75 or so records, and then wrote the CSV file out to come back again later and do the others.

But now when I re-run it (from the state checked-in here) I only get a single record displayed in the interface for reconciliation, and then it gives me a blank screen (not even with the CSV tray), even though there are 173 actually fuzzed, and still to do:

screen shot 2015-12-03 at 11 48 58

screen shot 2015-12-03 at 11 49 04

zarino:
Placing the first vote raises a Javascript exception which blocks the next pairing from being shown. The exception is: `Uncaught TypeError: Cannot read property 'name' of undefined` and it's being raised from this line: https://github.com/everypolitician/everypolitician-data/blob/1b481ecc0b34b8693ea9bf52f257eba762a78fea/templates/reconciliation.html.erb#L93

zarino:
So it looks like, in this case, `person` is undefined. I wonder why…

zarino:
Here’s what’s happening:

1. The UI attempts to display incoming person `id:Q5995801` (Marcelo Forni), and his three "potential matches" from the existing people in the database.
2. The third suggested person, `uuid:d6066e44-5f0e-4cb5-8e1c-c2378b50903d` (Marcela Hernando Pérez) appears in `window.reconciled`, and is recorded as having being matched up with incoming person `id:Q5995515`.
3. `highlightExistingVotes()` attempts to find person `id:Q5995515` in `window.matches`, but fails.
4. This means `personThisPersonWasMatchedTo` remains `undefined`, and when the template attempts to pull the `.name` property out of it, an exception gets raised.

I think the problem is that `highlightExistingVotes()` assumes `window.matches` will contain the person it's looking for – but when you've partially categorised a legislature, and then come back to it later, `window.matches` *doesn’t* contain all the people you previously dealt with. I think?

zarino:
I've temporarily wrapped a try/catch around the `highlightExistingVotes()` function call, which will soak up the exception and let you carry on categorising (just without the "this person was already matched to…" warning).

tmtmtmtm:
@zarino: yes `window.matches` seems to be slightly oddly named, but it's the list of people still needing to be reconciled. On a fresh run, this will be everyone, but where there were already existing matches, those will be in `window.reconciled` instead — though only as ID pairs, so the data we need isn't currently available anywhere in that case. I've tweaked this to just show the ID in those cases, rather than the name, which means we'll at least get _some_ warning, and will look into fixing it more properly later (probably through https://github.com/everypolitician/everypolitician/issues/235)

PR #1851, opened 2015-12-04
Reconciler: show matching fields as green or red

tmtmtmtm:
When reconciling, show whether each field matches or not, to make it easier to scan for errors.

screen shot 2015-12-04 at 15 05 54

Closes https://github.com/everypolitician/everypolitician/issues/234


tmtmtmtm:
@zarino: hopefully I've done the right thing in the sass…

This was all a bit fiddly to get right, but once I worked out _how_ to do it, it was relatively simple — but a huge win when actually reconciling.

zarino:
Sass looks fine! :+1:

PR #1940, opened 2015-12-07
Ukraine: refresh data

everypoliticianbot:
Automated data refresh for Ukraine - Verkhovna_Rada

#### Output

```
Regenerating sources/morph/data.csv
Creating term-8.csv
Creating names.csv

```


henare:
This is mainly changes from https://github.com/openaustralia/ukraine_verkhovna_rada_deputies/commit/b485f878fd3e221195979b6826ca6eda772afc94 which changed the area for constituency MPs from the dry district number to the far more useful region (which also should align with [Open Civic Data](https://github.com/opencivicdata/ocd-division-ids/blob/master/identifiers/country-ua.csv) names).

tmtmtmtm:
Thanks @henare. Any chance of keeping those district numbers as `area_id`s?

henare:
@tmtmtmtm I can certainly do that. Do you think that's better than using the [OCD IDs](https://github.com/opencivicdata/ocd-division-ids/blob/master/identifiers/country-ua.csv)?

tmtmtmtm:
If you can add full-blown OCD division IDs down to the constituency level (AFAICS the official repo only has region level atm) that would be ideal, but if that's going to take a while, then at least using the district numbers would be useful for now, and should presumably be quite a quick change?

henare:
Cool, that makes sense. I'll add the district numbers as `area_id`s now :+1:

henare:
I've made the change in https://github.com/openaustralia/ukraine_verkhovna_rada_deputies/commit/936c2fdb96de84a4f612343ec135e3536ab4882c and it's been rebuilt as #1944.

tmtmtmtm:
Thanks @henare  — merged that one.

henare:
Wonderful, thanks for your help as always @tmtmtmtm :octopus:

PR #2704, opened 2015-12-31
Peru: Additional Wikidata matches

tmtmtmtm:
The scraper now also picks up people on the English Wikipedia, from the template at https://en.wikipedia.org/wiki/Template:Peruvian_Congress_2011-2016


tmtmtmtm:
picked up automatically by later change

PR #3024, opened 2016-01-11
Extract CSV logic to everypolitician-popolo

chrismytton:
Part of https://github.com/everypolitician/everypolitician/issues/92
Part of https://github.com/everypolitician/everypolitician/issues/245


tmtmtmtm:
@chris I bumped the version of `csv_to_popolo`, and tweaked the call to `popolo.persons` in `generate_final_csvs` to be outside the loop rather than inside (as it's hugely expensive, and running it 11,000 times for the German Bundestag isn't particularly fast — I aborted that one after the first hour…) 

This seems to work cleanly everywhere now, but to avoid conflicts, I'll merge this without the data updates, and let those flow through from the normal rebuilder flow.

PR #3104, opened 2016-01-13
Reconciler: Add "Show me this card again later" option

chrismytton:
screen shot 2016-01-14 at 09 53 53

When there is an ambiguous match you might want to defer it until after
you've done the rest of the matching.

This change adds an option which can be activated with the '?' key, when
used this puts the card at the back of the stack so that it will be
shown again at the end of reconciliation.

Fixes https://github.com/everypolitician/everypolitician/issues/237

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -50,6 +50,9 @@
                     <div class="skip">
                       <header class="person__meta">None of these match</header>
                     </div>
+                    <div class="show-later">
+                      <header class="person__meta">Show me this card again later.</header>
+                    </div>
                 </div>
             </div>
         </div>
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
Looks great — though there's one odd edge-case here, where if you take the 'show me this again later' option on the last card, we hit the "we're finished" page before adding that person back onto the stack, meaning you never get them again, and they're left un-reconciled. At a quick glance that seems like it could be quite fiddly to handle, so it might be best merging this and making a new ticket to resolve that later — unless you can think of a simple way to deal with that?

chrismytton:
Hmm, can't think of a quick fix for that so I think probably best to merge this and ticket that separately.

tmtmtmtm:
:+1:

PR #3105, opened 2016-01-13
combine_sources: Speed up remap function

chrismytton:
Wrapping the call to 'combine_sources' with stackprof revealed that this
was the most expensive method being called. This change optimizes it to
use a hash lookup, which seems to speed up rebuilding a country quite
significantly.

### DIFF
--- a/rake_helpers/combine_sources.rb
+++ b/rake_helpers/combine_sources.rb
@@ -151,11 +151,10 @@ namespace :merge_sources do
     start_date: %w(start started from since),
     term: %w(legislative_period),
     website: %w(homepage href url site),
-  }
-  REMAP_LOOKUP = {}
-  REMAP.each { |key, values| values.each { |value| REMAP_LOOKUP[value] = key } }
+  }.each_with_object({}) { |(k, vs), mapped| vs.each { |v| mapped[v] = k } }
+
   def remap(str)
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
Looks good, though initialising a variable to nothing, and then immediately filling it in always makes me a little twitchy. I don't know a better way to do it in one pass in Ruby though, other than something like 

```
REMAP_LOOKUP = Hash[ REMAP.map { |key, values| values.map { |value| [value, key] } }.flatten(1) ]
```

Do you know of anything nicer than that?

chrismytton:
Perhaps [`Enumerable#each_with_object`](http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-each_with_object) would be a bit clearer?

```ruby
REMAP_LOOKUP =  REMAP.each_with_object({}) { |(key, values), remapped| values.each { |value| remapped[value] = key } }
```

tmtmtmtm:
Yep, that definitely feels more ruby-ish to me. Adding that also means we can get rid of needing to have the other temporary `REMAP` too, so I've added that in-place on to the end of creating the data structure. 

It might be heading towards being _too_ terse, when we factor this all out of Rakefiles we can tidy some of this up more.

PR #3336, opened 2016-01-20
Ukraine: Merge Wikdiata

tmtmtmtm:
Scraper at https://morph.io/tmtmtmtm/ukraine-deputies-wikidata (via https://morph.io/tmtmtmtm/ukraine-deputies-wikipedia, which gets data from https://uk.wikipedia.org/wiki/Верховна_Рада_України_VIII_скликання)

Closes https://github.com/everypolitician/everypolitician-data/issues/792


tmtmtmtm:
@henare Heads-up before I merge this, in case it causes any problems at your end. I think it should only _add_ data, rather than changing anything that was there before, but there's so much new data that github can't handle the diff very well, and it's vaguely possible that I've overlooked a material change.

There are also 8 people that I haven't been able to merge yet:
*   Missing: Кісельов Андрій Миколайович
*   Missing: Юзькова Тетяна Леонідівна
*  Missing: Онуфрик Богдан Семенович
*  Missing: Німченко Василь Іванович
*  Missing: Ревега Олександр Васильович
*  Missing: Підберезняк Вадим Іванович
*  Missing: Недава Олег Анатолійович
*  Missing: Логвинський Георгій Володимирович

The first few I've checked are definitely on the Wikipedia page we're using for source, and are getting picked up the scraper for that page. But the Wikidata lookup is missing them somehow. I'll investigate those ones separately later. But this gives us multi-lingual names for lots of people, gender for almost everyone, date of birth for 70% of them, and a whole bunch of other useful stuff.

tmtmtmtm:
Ah, it suddenly dawned on me what the problem is likely to be, and checking it shows I guessed right. The people who are missing don't actually _have_ Wikidata pages yet. I suspect that a bot added every page that was already in the Ukrainian Wikipedia at some point, but that that doesn't _keep_ them in sync, so new pages added on Wikipedia don't automatically get new Wikidata pages. So to get these matched up someone will need to create the pages in Wikidata and tie them back to their Ukrainian Wikipedia pages (where hopefully a suite of other bots will then add other handy cross-links etc), and we'll automatically pick up any future changes make there.

I hit this after the recent Canadian election too, where all the new people have had quite good Wikipedia pages created, but no Wikidata pages yet. I've been adding one or two a day of those, and other people seem to be quite good at extending the data in them once they exist, so if I take the same approach with the missing Ukrainian ones too, hopefully people will update those too.

tmtmtmtm:
@henare I'll probably merge this tomorrow, unless you raise any issues with it.

henare:
@tmtmtmtm many thanks for the heads up! I've loaded the data into a development version of Вони голосують для тебе and it seems to work fine so you should be good to merge (cc @beastie87).

PR #3405, opened 2016-01-22
Refactor, tidy, and document the Reconciliation code

tmtmtmtm:
As a lead-in to changing some of how Reconciliation works, tidy lots of the existing code up.


chrismytton:
Yep, this has certainly made a lot of the core concepts of matching/reconciliation much clearer. :+1:

PR #3443, opened 2016-01-23
Make generating the Reconciliation Interface an optional step

tmtmtmtm:
only generate the Reconciliation Interface if GENERATE_RECONCILIATION_INTERFACE is set.

Closes https://github.com/everypolitician/everypolitician/issues/277

With this, building France without a reconciliation interface now takes 32 seconds on my laptop, compared to 8 minutes previously.

### DIFF
--- a/lib/reconciliation/interface.rb
+++ b/lib/reconciliation/interface.rb
@@ -3,11 +3,13 @@ module Reconciliation
   class Interface
     attr_reader :merged_rows
     attr_reader :incoming_data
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
ping @chrismytton :roller_coaster:

chrismytton:
> With this, building France without a reconciliation interface now takes 32 seconds on my laptop, compared to 8 minutes previously.

Yay, that's quite a significant improvement! :+1:

PR #3453, opened 2016-01-23
Simplify the Reconciler

tmtmtmtm:
The Reconciliation Interface was responsible for too many things, making it difficult to reuse it in other places (e.g. for Membership as well as Person reconciliation). Simplify it.


tmtmtmtm:
cc @chrismytton

chrismytton:
:sparkles: :+1: :sparkles:

PR #3506, opened 2016-01-25
Reconciler: Show full field name when hovering

chrismytton:
Previously when a field name was truncated there was no way to see what the full field name was. This change adds a `title` tag to the `<dt>` so that the full field name appears on hover.

screen shot 2016-01-25 at 10 14 49

Fixes https://github.com/everypolitician/everypolitician/issues/295


tmtmtmtm:
Woot! Thanks. :rocket:

PR #3599, opened 2016-01-27
Ukraine: Add Wikidata party info

tmtmtmtm:
Add party information from Wikidata


tmtmtmtm:
@henare: another heads up — this time of extra information about the factions. I'm more confident that this one is OK, except for one issue. In the current data you have two different factions:

* 2846,"Група ""Партія ""Відродження"""
* 2795,"Група ""Відродження"""

My understanding is that this is probably to do with a merger (https://en.wikipedia.org/wiki/Revival_(Ukraine)#Merger) that I don't quite understand, and that I'm not sure is well reflected in Wikidata/Wikipedia. If it's important to you that these are treated separately, I can Do The Right Thing on Wikidata to make it all work cleanly, but I'd need someone to explain it to me better than that Wikipedia page currently does. 

Or, if it's just an artefact of the way you built things up, and actually they can be treated as one, then merging them is OK at this end too.

henare:
>I'd need someone to explain it to me better than that Wikipedia page currently does.

@lisoffsky @mmitre @beastie87 Can someone help with Tony's question?

lisoffsky:
So, there are some different issues here. 
Firstly its not a faction, its group. MPs can be in factions, groups or non-faction. Faction can influence on agenda and send people to the committees in priority (tbc). groups have less rights. 
Second. after elections some MPs united into group "Група ""Відродження""" but later they created same called party and became "Група ""Партія ""Відродження""" what means the word Party appeared in their title.

I think there is small difference in MPs who participated in those groups. 

Is this enough? Or lets dig in, just give a specific questions :)

tmtmtmtm:
Thanks @lisoffsky 

From what I can see everyone who was in "Група ""Відродження""" became part of "Група ""Партія ""Відродження""", on 2015-06-17. Then, on 2015-11-06, they were also joined by Ланьо Михайло Іванович.

I'm sure there's lots I don't quite understand about the distinction between groups and factions, as it's not one I remember coming across in any other country, but I think the key question for me is whether these are essentially two completely different things, or is more that Група "Відродження"	was either just _renamed_ as Група ""Партія ""Відродження"", or explicitly _replaced_ by it. Or to put it another way, would it still be possible for some people to be part of Група "Відродження" now, whilst others are part of Група ""Партія ""Відродження""?

lisoffsky:
I'm sorry for being a nerd.
Yes, you are totally right that's one thing. one group disappeared and the same day appeared with a bit different name. 

You are doing great job!

tmtmtmtm:
Hah, no worries. I'm completely fascinated by all these different systems, but with EveryPolitician we're trying to walk that delicate line between having sufficient detail so that someone who know all the nuances of a specific country's system has everything they need, but also not _requiring_ someone to know all that just to be able to use the information for more basic things.

Was the switch from Група "Економічний розвиток" to  Група ""Відродження"" also effectively just a renaming, or was that one more complex?

lisoffsky:
Let me say that Група "Економічний розвиток" migrated to Група "Воля народу".

tmtmtmtm:
Oh, that's even more complicated then. So when they became Група "Воля народу", all their members (other than Балицький Євген Віталійович ) left and joined Група "Відродження" instead?

tmtmtmtm:
I'm going to merge this so that things don't get blocked behind it. I think the data we have here is Good Enough for now, and I'll tweak the details of the mergers / renamings etc later.

PR #4007, opened 2016-02-07
Add a `popolo_url` field for each legislature

tmtmtmtm:
Rather than consuming applications needing to generate this themselves,
provide a URL straight into rawgit.com

Closes https://github.com/everypolitician/everypolitician/issues/244

(rebuilt `countries.json` not included so as to avoid clashes — once this is merged future PRs will automatically get the new format)

screen shot 2016-02-07 at 15 18 21

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -64,6 +64,7 @@ task 'countries.json' do
         json_file = h + '/ep-popolo-v1.0.json'
         name_file = h + '/names.csv'
         popolo = json_from(json_file)
+        remote_source = 'https://cdn.rawgit.com/everypolitician/everypolitician-data/%s/%s'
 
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
@chrismytton this looks like the most useful approach to me; or have you had any further thoughts on a better/different way to go about this?

chrismytton:
@tmtmtmtm Yep, I think this is going to be the simplest way to do this. It also matches the API in everypolitician-ruby - [`Everypolitician::Legislature#popolo_url`](https://github.com/everypolitician/everypolitician-ruby/blob/2867f271808c70ed65816efb957340ac6135991a/lib/everypolitician.rb#L105-L108). So :+1: from me!

PR #4832, opened 2016-02-28
Wikidata positions

tmtmtmtm:
Store P39 (position held) data for all people with Wikidata links.

To add these, from the directory of a legislature run:

1. `bundle exec rake build_p39s` — adds instructions for how to build
2. `bundle exec rake clean default` — build the legislature as normal
3. edit `sources/manual/position-filter.json` — move the entries from 'unknown' to 'exclude' or 'include'
4. `bundle exec rake clean default` — rebuild the legislature

### DIFF
--- a/rake_helpers/generate_final_csvs.rb
+++ b/rake_helpers/generate_final_csvs.rb
@@ -86,9 +86,13 @@ namespace :term_csvs do
     warn "Creating #{position_file}"
 
     positions = JSON.parse(File.read('sources/wikidata/positions.json'), symbolize_names: true) 
-    filter    = File.exist?(filter_file) ? JSON.parse(File.read(filter_file), symbolize_names: true).each do |s, fs|
-      fs.each { |f| f.delete :count }
-    end : { exclude: [], include: [] }
+    filter    = if File.exist?(filter_file) 
+      JSON.parse(File.read(filter_file), symbolize_names: true).each do |s, fs|
+        fs.each { |f| f.delete :count }
+      end
+    else 
+      { exclude: [], include: [] }
+    end
     to_include = filter[:include].map { |e| e[:id] }.to_set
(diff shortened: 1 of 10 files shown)


tmtmtmtm:
@chrismytton want to give this a quick sanity check before I merge it? For now it just spits out a CSV into `unstable/` rather than putting anything into the Popolo. I'd like to start gathering data for everywhere first (and sorting through which positions we actually want to include etc) before working out how to best represent it all in Popolo.

rake_helpers/generate_final_csvs.rb
```diff
@@ -78,4 +78,47 @@
...
+  task :positions => ['sources/wikidata/positions.json', 'ep-popolo-v1.0.json'] do
+
+    filter_file   = 'sources/manual/position-filter.json'
+    position_file = "unstable/positions.csv"
+    warn "Creating #{position_file}"
+
+    positions = JSON.parse(File.read('sources/wikidata/positions.json'), symbolize_names: true) 
+    filter    = File.exist?(filter_file) ? JSON.parse(File.read(filter_file), symbolize_names: true).each do |s, fs|
+      fs.each { |f| f.delete :count }
+    end : { exclude: [], include: [] }
```
chrismytton: I think these lines might be clearer as an if statement, the ternary operator is a bit hidden otherwise!

```ruby
filter = if File.exist?(filter_file)
  JSON.parse(File.read(filter_file), symbolize_names: true).each do |s, fs|
    fs.each { |f| f.delete :count }
  end
else
  { exclude: [], include: [] }
end
```

chrismytton:
Other than the comment I made this all looks good to me! :+1:

rake_helpers/generate_final_csvs.rb
tmtmtmtm: Yep. Fixed.

PR #4903, opened 2016-03-01
Position Files: arrange the Filter file by Position type

tmtmtmtm:
Whilst we're building the Filter files, we should group the Positions by
what type they are — e.g. the Membership of this or other legislatures;
memberships of the Executive, etc.

This will let us process them differently, when we get around to doing more
(or output them better in Popolo)


tmtmtmtm:
@chrismytton mostly as an FYI for you…

PR #5196, opened 2016-03-05
Namibia (Council): refresh for new 2015-2020 term

tmtmtmtm:
Data after 2015 election


tmtmtmtm:
@duncanparkes Your scraper for this looks like it has generated lots of almost-duplicate entries. My guess is that the names on the site changed at one point, and because you're just slugifying them to create an ID, they've ended up with new entries after that.

It may be enough to simply empty the database and re-run, though it seems like it would be better to use the people IDs from their URLs

duncanparkes:
I'm afraid neither the id number, nor the slug are usable. These urls are all for the same person:

http://www.parliament.na/index.php?option=com_contact&view=contact&id=1317:nicanor-hilma-ndinelago&catid=136&Itemid=1376
http://www.parliament.na/index.php?option=com_contact&view=contact&id=29:nicanor-hilma-ndinelago&catid=108:all-members&Itemid=1376
http://www.parliament.na/index.php?option=com_contact&view=contact&id=915:nicanor-hilma-ndinelago-3rd&catid=124&Itemid=1377

Perhaps we might be able to use the slug part of the id, with the 3rd (or whatever) removed from th end. Otherwise, it might be empty the DB and re-run, I'm afraid.

tmtmtmtm:
Ouch, that's nasty. Looks like we might need to manually reconcile them term by term.

duncanparkes:
It might not be as bad as that. I'm going to try taking the id from the query, removing the number and colon from the beginning, and removing things that look like '-3rd' etc from the end, and see if *those* look like they are unique.

I have got back to understanding why I slugified the names!

duncanparkes:
Removing '3rd' etc, doesn't fix it. We still have 'margret-mensah' and 'margretmensah'. I wonder if the email addresses are unique and always exist.

duncanparkes:
Answering my own question - the fact there's code to cope with them being missing suggests not. I can't see a better ID anywhere, I'm afraid.

duncanparkes:
@tmtmtmtm I'm going to need you to tell me the most useful thing I can do for you here.

tmtmtmtm:
I think it's going to be term-by-term reconciliation then. I doubt I'll get time to do that until next week though.

duncanparkes:
What's the best thing for me to set as an ID then to make that easiest? Is there anything that's an improvement on what we have now?

tmtmtmtm:
As long as there's never a case where two different people have the same ID, I can handle the cases where one person has two (or more) different IDs.

duncanparkes:
In that case, perhaps we should use the id query parameter somehow. The number part seems pretty useless, but hopefully the slug part is unique to a person. I'll change the IDs to use the part of that ID query string. Would you prefer the term suffix kept in or removed?

tmtmtmtm:
We'll be throwing these IDs away, so what they look like is pretty much irrelevant. It would be useful, though, if the people from the earliest term kept the same IDs as they originally had, as then we can build up the successive data from there, rather than having to start from scratch again and manually make sure that people get given the same UUIDs.

tmtmtmtm:
added in https://github.com/everypolitician/everypolitician-data/pull/12190

PR #5440, opened 2016-03-08
Nigeria (Senate): initial data

tmtmtmtm:
Data extracted from ShineYourEye


chrismytton:
Not sure what's going on but there seem to be lots of "Peoples Democratic Party" orgs rather than just one - https://github.com/everypolitician/everypolitician-data/blob/23d76258af50e6e9e3f2c0978353b7e8a8276335/data/Nigeria/Senate/ep-popolo-v1.0.json#L1889-L2258.

Looks like this is the case for other parties as well.

tmtmtmtm:
@chrismytton, this is because they all have different `party_id` entries in the source https://github.com/theyworkforyou/shineyoureye_data/blob/a0235e9c3ad57a26e4046ea2f5e5e5a8be9371ac/senate_legislative_memberships.csv

tmtmtmtm:
@chrismytton is this PR still active?

PR #6094, opened 2016-03-18
Reconciler: warn if there are multiple IDs or UUIDs after matching

mhl:
At the end of matching all incoming people to existing people, it's
possible that the matches between IDs (of incoming people) to UUIDs
(of existing people) aren't one-to-one.  With the change in this
commit, a warning will be displayed at the end if this is the case,
including the IDs / UUIDs so this can be investigated further.

reconciler-duplicate-ids

(n.b. I slightly fiddled the data to make that screenshot, so the actual
IDs won't make sense...)

Fixes https://github.com/everypolitician/everypolitician/issues/240

### DIFF
--- a/templates/reconciliation.js
+++ b/templates/reconciliation.js
@@ -239,14 +239,14 @@ jQuery(function($) {
     var incomingPerson = match.incoming;
 
     // If there's one and only one 100% match, choose it automatically
-    var exactMatches  = _.filter(match.existing, function(e) {
-      return e[1] == 1;
-    });
-    if (exactMatches.length == 1) {
-      window.autovotes.push( [incomingPerson.id, exactMatches[0][0].uuid] );
-      return;
-    }
-
+    // var exactMatches  = _.filter(match.existing, function(e) { 
+      // return e[1] == 1;
+    // });
+    // if (exactMatches.length == 1) { 
+      // window.autovotes.push( [incomingPerson.id, exactMatches[0][0].uuid] );
+      // return;
+    // }
+    
     var incomingPersonFields = _.filter(Object.keys(incomingPerson), function(field) {
       return incomingPerson[field];
     });
--- a/templates/reconciliation.js
+++ b/templates/reconciliation.js
@@ -20,7 +20,7 @@ var vote = function vote($choice){
   var vote = [];
 
   if($choice.is('.skip')) {
-    // Insert a null value to indicate skip. 
+    // Insert a null value to indicate skip.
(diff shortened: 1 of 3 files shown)


templates/reconciliation.js
```diff
@@ -35,6 +35,48 @@ var vote = function vote($choice){
   nextPairing($pairing);
 }
 
+// This function looks through allVotes() to see if there are any IDs
+// matched with multiple UUIDs or (conversely) any UUIDs matched with
+// multiple IDs. If so, it returns HTML that warns about these
```
tmtmtmtm: It would be good to split this in two, so that the HTML generation is separate from finding the problem cases. Simply showing the HTML output is a good first step, but to really be useful we'll need to offer an interface to _do_ something with these (even if just to un-match them for later re-matching), at which point we'll need the data, not just the text.

templates/reconciliation.js
```diff
@@ -42,7 +84,9 @@ var nextPairing = function nextPairing($currentPairing){
     highlightExistingVotes($nextPairing);
     $nextPairing.show();
   } else {
-    $('.messages').html('<h1>Reconciliation complete!</h1>'); 
+    $('.messages').html(
+      '<h1>Reconciliation complete!</h1>' + getDuplicateIDsHTML()
+    );
```
tmtmtmtm: It might be better to display this any time the CSV tray is opened, not only when you get to the very end. Often, when reconciling a lot of data, I'll do the reconciliation in multiple phases, and it would be great to be warned of duplicates at any time I 'finish' rather than only when I'm completely done.

templates/reconciliation.js
mhl: Yeah, I noticed after making the PR that the description in the spreadsheet actually asks for much more work than the GitHub issue (i.e. it also wants an interface for unmatching or representing those matches). I can certainly split out a function to return duplicates separate from the HTML generation, but I'd be tempted to create a new issue for the more involved work of offering an interface for fixing them. (And possibly only work on that after introducing some tests for the reconciler - doing anything more complex without them would feel brittle, since it's awkward to test by hand anyway.)

tmtmtmtm: Oh, I certainly meant that the interface would be a new thing — just trying to make sure that that can be purely additive, rather than having to also split things like this up.

mhl:
@tmtmtmtm I've pushed a new version which addresses those comments, hopefully.

tmtmtmtm:
Thanks @mhl

PR #6099, opened 2016-03-18
Switch urls in countries.json to use GitHub directly

chrismytton:
RawGit has been down a couple of times lately, so instead of pointing to
that this changes things to point directly to GitHub. This does mean
that all JSON and CSV urls will have the content type text/plain (as
that's what GitHub serves them up with) but we can document that and
tell people to use RawGit if they need the correct content type.

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -67,7 +67,7 @@ task 'countries.json' do
       legislatures: hs.map { |h|
         json_file = h + '/ep-popolo-v1.0.json'
         name_file = h + '/names.csv'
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
I'm not sure we can just change these without better notification etc. There could easily be webapps using this that require these to be URLs that get the correct headers, etc.

We probably either need to manually munge the URL in our consuming sites, or add a parallel structure here.

tmtmtmtm:
I'm going to close this. It wasn't quite the right approach, and hasn't been needed since.

PR #6836, opened 2016-03-29
Reconciler: Some refactoring

tmtmtmtm:
The logic for showing the cards on the left and right side (incoming vs
existing people) was getting a little hard to follow, as they were both
using the same template, even though they were conceptually quite
different things, and almost every section of the card had to be wrapped
in a check to see which type it was.

Split these out so that it's easier to follow (even though there’s now a
small amount of duplication between them), but also harmonise some of
the other code that was needlessly different.

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -58,7 +58,61 @@
         </div>
     </script>
 
-    <script type="text/html" id="person">
+    <script type="text/html" id="incomingPerson">
+      {% if(person.uuid){ %}
+        <div class="person" data-uuid="{{ person.uuid }}">
+      {% } else if(person.id){ %}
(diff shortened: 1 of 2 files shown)


chrismytton:
Yep, this is much easier to follow with these two template split up. I think it was probably the case that they were very similar to start with and then drifted apart. :+1: :ship:

tmtmtmtm:
Yeah — that's a pattern I've noticed a few times now. We start out doing a very simple version of something, and end up with two things that look close to identical, so abstract them out into a common form. But then as we add more and more code later it becomes increasingly obvious that the things were actually quite different — but there's never a key point where it's obvious enough that we should split them up rather than adding just a single additional `if`…

PR #7013, opened 2016-04-01
Store aliases for Groups

tmtmtmtm:
As well as getting labels from Wikidata for Groups (and Areas etc), also
store their aliases.


chrismytton:
:+1:

PR #7014, opened 2016-04-01
Add Membership Matrix source

chrismytton:
Pulls in data from a Membership Matrix CSV and uses it as the main source of truth for memberships in a country if that source is specified.

Part of https://github.com/everypolitician/everypolitician/issues/373


rakefile_common.rb
```diff
@@ -156,6 +156,17 @@ def instructions(key)
...
 end
 
+def sources
+  @sources ||= begin
+    # Make sure all instructions have a `type`
+    if (no_type = instructions(:sources).find { |src| src[:type].to_s.empty? })
+      raise "Missing `type` in #{no_type} file"
+    end
+
+    instructions(:sources).map { |s| Source::Base.instantiate(s) }
```
tmtmtmtm: It would probably be cleaner to do the check for a missing `type` inside this `map`, rather than as a separate scan beforehand.

chrismytton: Ah yes, we did discuss this but I forgot to make the change! Fixed in https://github.com/everypolitician/everypolitician-data/commit/3a15a9f8608030b7957918246a78b6911cc009b6.

chrismytton:
I've rebased this onto master so it will merge cleanly.

tmtmtmtm:
@chrismytton — is this ready to merge or did you have some other things you wanted to do to it first? (If it's the latter, it might be good to pull `9f331c2` out into a separate PR

chrismytton:
@tmtmtmtm This is ready to be merged now. The changes I need to make are in the membership matrix UI, which I will ticket separately.

rakefile_common.rb
```diff
@@ -156,6 +156,15 @@ def instructions(key)
   @instructions[key]
 end
 
+def sources
+  @sources ||= begin
```
tmtmtmtm: The extra level from this `begin` is a little surprising, and seems unnecessary. 

Wouldn't this be simpler as just:
```ruby
def sources
  @sources ||= instructions(:sources).map do |src|
    raise "Missing `type` field in source: #{src}" if src[:type].to_s.empty?
    Source::Base.instantiate(src)
  end
end
```

chrismytton: Oops, yep, fixed in https://github.com/everypolitician/everypolitician-data/pull/7014/commits/5b6c13bb6eb4f62ff4385fa819edd5b7f80eb8b5.

PR #7494, opened 2016-04-07
Uganda spreadsheet

chrismytton:
Uses the new Google Spreadsheet that Collin from Parliament Watch Uganda created.


tmtmtmtm:
This isn't completely finished yet, as we still need to include the other Special Interests members, but it's Good Enough For Now™

PR #7749, opened 2016-04-11
added gender

marfi:


tmtmtmtm:
@marfi this file is auto-generated every time we rebuild from the sources, so modifying it will be blown away pretty much instantly. 

I don't know of any official source for gender data from Croatia: almost all the information we have currently comes from http://Gender-Balance.org, so feel free to play there (and get some other people to play as well, as we need multiple 'votes' per person before we're confident enough to accept it) :)

Otherwise, if you really want to supply gender information for a single person like this, you'll need to create a 'corrections.csv' file, similar to the one for Finland: https://github.com/everypolitician/everypolitician-data/blob/master/data/Finland/Eduskunta/sources/manual/corrections.csv

(Apologies that this isn't very well documented)

marfi:
Thought so. Thanks for the additional info.

tmtmtmtm:
closed by https://github.com/everypolitician/everypolitician-data/pull/7866

PR #8712, opened 2016-04-25
Add bin/gender_information.rb script

chrismytton:
This calculates the gender information for politicians that we currently
have information for in the most recent term for each legislature.

Output at time of writing:

```
Female: 23.75% (5714)
Male: 76.25% (18350)
Total: 24064
```


tmtmtmtm:
I suspect you want this to be info**r**mation, rather than infomation…

But aside from that, was there a specific requirement for this?

chrismytton:
@tmtmtmtm Yep, it was so that we could reply to Stephen Abbott's tweet from this morning - https://twitter.com/stephen_abbott/status/724497369853755392

tmtmtmtm:
Not factoring in people for whom we don't _know_ the gender makes the results a little odd. As a one-off for that purpose it's probably OK, but I'm not sure it rises to the level of a script that we want to keep around in the repo — at least not without a lot of extra caveats / general comments as to purpose etc.

StephenAbbott:
Thanks for this, @chrismytton

chrismytton:
@tmtmtmtm Yep, fair points. Will close this PR as I think you're right, this doesn't belong in this repo.

@StephenAbbott You're welcome!

PR #9286, opened 2016-05-03
Require 'corrections' to be an explicit Source

tmtmtmtm:
Rather than looking for a "corrections" file at a specified location,
require it to be explicitly set in the `instructions.son` file to be used.


chrismytton:
👍  Looks good to me!

PR #9781, opened 2016-05-11
Update shebang lines in scripts to use /usr/bin/env ruby

mhl:
A shebang line which hardcodes /usr/bin/ruby may cause problems
if you are using rbenv or in any other situation where the version
of ruby that comes first in your PATH is not /usr/bin/ruby.
Using /usr/bin/env ruby in the shebang line will exec the ruby
binary that comes first in your PATH.

Investigated with Dave Whiteland (@davewhiteland).

### DIFF
--- a/bin/last_updated.rb
+++ b/bin/last_updated.rb
@@ -1,4 +1,4 @@
-#!/usr/bin/ruby 
+#!/usr/bin/env ruby
 
 # How long has it been since the data for a legislature changed?
 
--- a/bin/learn_position.rb
+++ b/bin/learn_position.rb
@@ -1,4 +1,4 @@
-#!/usr/bin/ruby 
+#!/usr/bin/env ruby
 
 require 'json'
 require 'pry'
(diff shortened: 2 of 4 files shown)


tmtmtmtm:
I'm curious as to what the trigger for this was. I'm assuming that the issue was only with one of these files, and the other two were simply fixed at the same time?

mhl:
Yes, that's right. Dave was trying to run `bundle exec bin/last_updated.rb` but even though he'd run `bundle install` successfully was getting the error that `addressable` couldn't be found. I tried to reproduce it and got an error that my Ruby version was too early, even though I'd set one permitted by the `Gemfile` using `rbenv`. Both of our problems were fixed by changing the shebang line to use `/usr/bin/env` instead. I changed the other two sheban lines on spec because having `/usr/bin/ruby` hardcoded is asking for similar problems in the future, I think.

tmtmtmtm:
Hmmm. Most of the files here don't even _have_ a shebang — I normally execute them all with an explicit `ruby` in the command, e.g. `bundle exec ruby bin/last_updated.rb`

Would that have also been problematic for you? Should we add a shebang like this to the start of **all** the scripts?

mhl:
For me having that shebang line in each script and having them executable would be helpful, yes. (I don't see why we wouldn't want to support them being invoked in that way.)

PR #9845, opened 2016-05-12
Reconcile Gerry Adams across IE, NI and UK

tmtmtmtm:
DO NOT MERGE — This is just proof of concept at the minute. We also want to store the remapping into another file, similar to the legacy IDs, so that it can be stored on `Person`s as a legacy ID (and then teach EveryPolitician::Popolo to seamlessly look up by those when a ID lookup fails)

Give Gerry Adams the same UUID in all three legislatures.


tmtmtmtm:
Closing as this was just a proof of concept.

PR #10104, opened 2016-05-16
Better error messages when fetching remote urls

chrismytton:
Rather than printing a big stack trace when an http request fails we instead just print an error message and exit.


tmtmtmtm:
👍

PR #10106, opened 2016-05-16
Switch the Madagascar parliament parser from Duncan -> Mark

mhl:
I've forked Duncan's scraper, and made a few changes (purging bad
data, changing to a better primary key) so it would make sense
to take data from my fork now.


tmtmtmtm:
@mhl I re-reconciled the data with the new IDs, and everything seems to have gone smoothly … other than that pretty much every membership seems to be for a different constituency now. Any ideas what's happened there?

tmtmtmtm:
@mhl The changing constituency information is because the order of columns in the scraper has changed significantly. [Previously](https://github.com/everypolitician/everypolitician-data/blob/a44f44a9f47666e0911e35239fe1413161dde529/data/Madagascar/Assembly/sources/morph/data.csv) the order was 

```
chamber,region,details_url,name,party,image,district,executive,term_id,term
```

but [now](https://github.com/everypolitician/everypolitician-data/blob/687a5eb9a12031cb8cd73d852d3e13862ee316c5/data/Madagascar/Assembly/sources/morph/data.csv) it's 

```
party,page_id,district,term_id,image,name,region,details_url,chamber,term
```

By default EP takes the last "area-like" column as the constituency: which in the earlier data was `district`, but is now `region`.

We should perhaps [look at combining both](https://github.com/everypolitician/everypolitician-data/issues/10264), but for now I've simply [changed the instructions](https://github.com/everypolitician/everypolitician-data/pull/10106/commits/e02546ee39d1b9b59e714be73d6f75eef4a00e4d) to explicitly pick up the `district` column instead.

tmtmtmtm:
I also noticed that the name of the current terms is [currently](https://github.com/mhl/madagascar/blob/2325cad07bd943ef059117669257e9ed9d4c4505/scraper.py#L83) being set in the scraper to "2013-1018", rather than 2018! 

As this is just hard-coded there, rather than actually scraped, there's not much point in pulling it from Morph, so I've just moved to a manually maintained `terms.csv` file instead, and corrected it locally (i.e. you could remove the final 5 lines of the scraper with no ill effect)

mhl:
@tmtmtmtm I've removed the saving of the hardcoded term data now.

In the scraper, the data for each member is stored in a dictionary (so unordered) and a list of members is passed to `scraperwiki.sqlite.save` - from the documentation [here](https://classic.scraperwiki.com/docs/python/python_datastore_guide/) and [here](https://github.com/scraperwiki/scraperwiki-python) it doesn't look as if there's an obvious way of controlling how the keys of those dictionaries are ordered when the library creates the schema in SQLite (unless you create the schema manually). So that might be something to watch out for with other scrapers - in this case the order would have changed, I suppose, because I created a new morph scraper and thus a new database.

tmtmtmtm:
Ah, good to know, thanks. Another good reason why the scrapers should all be in Ruby, then? 😜

PR #10165, opened 2016-05-17
Remove duplicate memberships from Popolo output

chrismytton:
If duplicate memberships are detected we emit a warning and then discard
the duplicates.

Fixes https://github.com/everypolitician/everypolitician/issues/395

### DIFF
--- a/rake_helpers/generate_ep_popolo.rb
+++ b/rake_helpers/generate_ep_popolo.rb
@@ -100,6 +100,11 @@ namespace :transform do
       m.delete :start_date if m[:start_date].to_s.empty? || (!e[:start_date].to_s.empty? && m[:start_date].to_s <= e[:start_date].to_s)
       m.delete :end_date   if m[:end_date].to_s.empty?   || (!e[:end_date].to_s.empty?   && m[:end_date].to_s   >= e[:end_date].to_s)
     end
+    duplicates = @json[:memberships].size - @json[:memberships].uniq.size
+    if duplicates > 0
+      warn "Discarding #{duplicates} duplicate membership#{duplicates != 1 ? 's' : ''} from Popolo".yellow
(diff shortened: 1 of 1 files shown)


rake_helpers/generate_ep_popolo.rb
```diff
@@ -100,6 +100,11 @@ def terms_from_csv
       m.delete :start_date if m[:start_date].to_s.empty? || (!e[:start_date].to_s.empty? && m[:start_date].to_s <= e[:start_date].to_s)
       m.delete :end_date   if m[:end_date].to_s.empty?   || (!e[:end_date].to_s.empty?   && m[:end_date].to_s   >= e[:end_date].to_s)
     end
+    duplicates = @json[:memberships].size - @json[:memberships].uniq.size
+    if duplicates > 0
+      warn "Discarding #{duplicates} duplicate membership#{duplicates != 1 ? 's' : ''} from Popolo".yellow
```
tmtmtmtm: This would be a lot better if if reported something about what the duplicates were, so they can actually be investigated…

tmtmtmtm: Getting a list of the duplicates is a little trickier than I'd hoped as `@json[:memberships] - @json[:memberships].uniq` ends up empty (as `[1, 2, 2, 3] - [1, 2]` = `[3]`, not `[2, 3]` )

Best I could come up with on a very quick play was:

`@json[:memberships].group_by { |m| m }.select { |_, ms| ms.size > 1 }.map(&:first)`

chrismytton: Done in 2031e73bda1db025d573c7dde2989a13952d15f9

tmtmtmtm:
Running this against the entire dataset, it seems we're going to catch quite a few of these…

PR #10174, opened 2016-05-17
UK (Commons): Twitter list source

chrismytton:
Uses the data from https://morph.io/everypolitician-scrapers/twitter-gov-uk-mps which in turn scrapes https://twitter.com/gov/lists/uk-mps to get information from Twitter.


tmtmtmtm:
@chrismytton I've pushed a tweak to remove the "MP" suffix from people's names, as I don't think it's that useful to have those all flow through as alternatives. 

I'm in two minds whether it's better to even just drop the name field entirely. We get a couple of useful additional names from this, but we also pick up a lot of fairly useless ones…

PR #10227, opened 2016-05-18
Bangladesh (House): Update source

chrismytton:
Switch to https://morph.io/everypolitician-scrapers/bangladesh-parliament-scraper which is the new version of the scraper.


chrismytton:
OK, this is *finally* pulling in the data from the new scraper as well as preserving the people from the 9th term that have disappeared from the official site.

PR #10229, opened 2016-05-18
UK (Commons): switch priority order for Twitter handles

tmtmtmtm:
Bump up the priority of twitter handles from the Twitter List, but keep the priority of Twitter images low.


chrismytton:
LGTM! 👍

PR #10299, opened 2016-05-19
Build speed: Pre-generate date_headers list

tmtmtmtm:
Rather than asking each row of `merged.csv` in turn what date columns it
has, pre-build the list of all the date columns in the CSV.

In the most extreme case (Turkey, with over 12,000 rows) this speeds
this task from ~40 seconds to ~0.2 seconds!


tmtmtmtm:
@chrismytton FYI (ouch!)

PR #10344, opened 2016-05-20
Use rcsv for parsing CSV sources

chrismytton:
Switch to using [rcsv](https://github.com/fiksu/rcsv) to speed up reading the various CSV files that we build our data from.


tmtmtmtm:
@chrismytton This breaks quite badly as it tries to convert the `uuid` column from GenderBalance to an `int`, effectively losing all the gender data.

PR #10599, opened 2016-05-24
Rebuild a single entry in countries.json

tmtmtmtm:
Rebuilding the entire countries.json from scratch every time is very
slow. So allow an optional option to only rebuild a single country.

This still almost certainly isn't the best approach to this — we should
probably explicitly keep this updated, file by file, based on commits
— but it's an improvement over what's currently there, at least for
developing locally. We can work out how to improve the interaction with
the bot later.

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -50,15 +50,25 @@ end
 
 desc 'Install country-list locally'
 task 'countries.json' do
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
👍  from @chrismytton in Slack

tmtmtmtm:
Hmmm. Just realised that this will blow up if it's trying to add a new country that isn't already in the file…

PR #10632, opened 2016-05-24
Switch to SPARQL endpoint for election lookup

chrismytton:
The WDQ url isn't very reliable so this changes the election lookup to use the SPARQL query endpoint instead.

First part of https://github.com/everypolitician/everypolitician/issues/402

### DIFF
--- a/lib/wikidata_lookup.rb
+++ b/lib/wikidata_lookup.rb
@@ -126,7 +126,7 @@ class ElectionLookup < WikidataLookup
   # We don't have the normal id => uuid Hash here, 
   # but rather instructions for a WDQ lookup
   def initialize(instructions)
-    q = 'CLAIM[31:%s]' % instructions[:base].sub(/^Q/,'')
+    q = "SELECT ?item WHERE { ?item wdt:P31 wd:#{instructions[:base]} . }"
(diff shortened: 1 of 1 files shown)


lib/wikidata_lookup.rb
```diff
@@ -148,12 +148,12 @@ def other_fields_for(result)
   end
 
   private
-  WDQ_URL = 'https://wdq.wmflabs.org/api'
+  WDQ_URL = 'https://query.wikidata.org/sparql'
 
   def wdq(query)
```
tmtmtmtm: probably better to rename the method and URL etc, as WDQ is the name of the specific Wikidata tool, rather than just a generic concept, so this could be quite confusing later.

tmtmtmtm:
That was simpler than I had feared! Looks good, other than suggestion to rename things to avoid future confusion.

lib/wikidata_lookup.rb
chrismytton: Done in https://github.com/everypolitician/everypolitician-data/commit/fc614096d84ef5d3186099289c8dc3c42de33689

tmtmtmtm:
👍

PR #11200, opened 2016-05-31
Convert event ids to strings when sorting

chrismytton:
Some events have string ids and others have symbols. When trying to sort
events those that have the same start_date fall back to sorting by id,
which was failing because of the comparison between a symbol and a
string.

This should fix most of the Rollbar errors that have come in over the past couple of days.


tmtmtmtm:
👍

PR #11321, opened 2016-06-01
Refactor area lookup

chrismytton:
This takes the lookup logic that was specific to Australia and repackages it to be more generic, in this case so we can use it with Uganda.


tmtmtmtm:
This is changing the behaviour of the Australian Representatives. It's throwing out lots of warnings (many of which are duplicates), but also looks like it's getting very confused by the Northern Territory / Solomon split (https://en.wikipedia.org/wiki/Division_of_Northern_Territory), as it's changed records in both directions on that. This may be because our data is wrong somewhere (and if so it's good that this has highlighted it), but simply changing behaviour for the only country it previously worked for probably isn't good, unless the current behaviour is broken at this is actually fixing that (in which case that should be noted explicitly.)

Whilst this is all still in your head, it's probably worth also documenting some of it somewhere (at a minimum, just in comments, though ideally somewhere better, e.g. a nascent guide to `instructions.json`). The removal of `"generate": "area"` is a little opaque at the minute (and yes, this one is definitely my fault in the first place!)

chrismytton:
I've fixed the Australia problem, I had missed out the logic for dealing with the `overrides` section of `instructions.json`, that's working correctly now. I've also changed it so it only outputs warnings once for each area, as it was previously.

I've also added some *very* basic documentation for the OCD identifiers.

tmtmtmtm:
I'm a little worried that this reinforces the default approach being a fuzzy search, rather than an exact one (with an option for fuzziness). I don't really want that to be a block to getting this live, but if there's something quick you can do whilst this is all still in your head, then it might be good. Otherwise maybe just a matter of documenting that part and iterating later.

chrismytton:
> I'm a little worried that this reinforces the default approach being a fuzzy search, rather than an exact one (with an option for fuzziness). I don't really want that to be a block to getting this live, but if there's something quick you can do whilst this is all still in your head, then it might be good. Otherwise maybe just a matter of documenting that part and iterating later.

Sorry, missed this comment on Friday! I've changed things so that it now defaults to an exact match unless you explicitly specify `"fuzzy": true` in `instructions.json`. I've updated Australia and Uganda to use fuzzy matches so they work correctly still.

tmtmtmtm:
Hmm. That `fuzzy` option seems like it should go into the `merge` hash with the `overrides` etc., rather than being at the top level of the hash. Though we already broke the pattern slightly with the `generate: area` option for OCD files previously. 

I think we're falling victim a little here to the concepts instructions files not being clearly documented and/or some of these categories not being very well named. But when we _do_ document them better, I think I'd prefer to really have a smaller number of high level sections that apply to processing pretty much every type of source: (`file`, `create`, `source`, `type`, `merge`, `notes`), with everything else within those, rather than lots of source-specific details at this level too.

chrismytton:
Hmm yeah, good point, we don't want the top level of the source object just becoming a dumping ground for a multitude of options. I think as you say, for now adding it into the `merge` key makes sense. I will make that change now.

chrismytton:
OK that's done now.

tmtmtmtm:
👍

PR #11829, opened 2016-06-08
Add a `stats.json` file

tmtmtmtm:
Rather than generating stats from a `bin/` file (as we currently do in
the dashboard script, and others), auto-generate them any time we're
rebuilding a legislature.

This will live in `unstable/` whilst we play around with the format etc for it.


chrismytton:
Looks good to me! 👍

PR #11835, opened 2016-06-08
Match multiple groups to a single wikidata identifier

chrismytton:
Allows multiple groups to be matched up to the same wikidata identifer. Previously this didn't work if multiple groups were matched up to a single wikidata id, in that case only one group would be correctly matched to wikidata.

Initial work towards https://github.com/everypolitician/everypolitician/issues/428


tmtmtmtm:
I added Antigua and Barbuda here as an example of this in action (including seeing the "Parties matched to Wikidata" stat going up :))

PR #11959, opened 2016-06-09
Northern Cyprus (Assembly): Merge birth dates from 'images' source

### LINKED ISSUE #11932: Northern Cyprus: add birth dates
The instructions file could be updated to pick up the birth dates (years) from https://morph.io/wfdd/nc-assembly-scraper.

wfdd:
Closes #11932.


wfdd:
I closed my earlier PR cuz I force-pushed after forgetting to clobber the data and I didn't wanna confuse the bot or whatnot.

tmtmtmtm:
Looks great, thanks! 

The only small suggestion I would have for future ones is to commit the files that you've hand-edited (e.g the `instructions.json` file) separately from the ones that get generated from the `rake` task (the way you did with the reconciliation file). That makes it easier to cherry-pick or undo etc later if needed.

wfdd:
It doesn't look like it's updated `countries.json` automatically by the way.

tmtmtmtm:
Yeah, I think that's a permissions thing. Because you're an "external" contributor, Github treats the Pull Requests quite differently from ones from someone on the team, and none of us (whether human or bot) have the ability to add new commits to them.

It's not a massive problem for the commit to go through without it, as the next time `countries.json` is built, it'll notice and correct, but it means there's a tiny period where the file is out of sync with the true state of things. Will need to give this a bit of thought to work out the best general solution.

wfdd:
I regenerated `countries.json` but committed just the Greenland 'hunk' in my other PR, which appears to have worked.  If automating the process for outside contributors is not an option then perhaps the `rake` command could be made to accept a legislature option (or, if invoked from within a legislature folder, it could automatically detect the fact).  But working in a 'shallow' state in git is a bit of a nuisance in general as I've come to find out.

tmtmtmtm:
We added a single-country rebuild option a few weeks ago. It's not documented anywhere other than [in comments](https://github.com/everypolitician/everypolitician-data/blob/f04f52b9768eadd1423da3ab79033b4b0edfc556/Rakefile.rb#L53-L57) yet, but you can do  `EP_COUNTRY_REFRESH=Greenland bundle exec rake countries.json`

wfdd:
Oh cool.  If it would also stage the file and automatically fill in the commit message for me, it'd be perfect ;)

tmtmtmtm:
That would be a little too tightly coupled to a particular workflow, I think. I have a shell alias that does all of that though :)

PR #12013, opened 2016-06-10
Fix click voting in reconciler

wfdd:
Previously, it would attempt to extract the UUID from the header
element though the UUID attribute is actually set on its parent.


tmtmtmtm:
Ouch, that's a nasty one. Presumably we've never noticed that before because we're all using the keyboard instead!

Thanks @wfdd

PR #12020, opened 2016-06-10
Australia (House of Representatives) collect twitter data from openaustralia scraper

davewhiteland:
Spotted that we're not pulling in the twitter data that the OpenAustralia scraper already has, so added `twitter` to the columns we're using.

(Right now, this doesn't pick up any new data, but the more sources the merrier)


tmtmtmtm:
👍 

@davewhiteland are you going to do an equivalent PR for the Senate too?

tmtmtmtm:
Actually I've reverted this, because it didn't include the actual rebuild of the data…

Issue #5773, opened 2016-03-13
Greenland: new terms


tmtmtmtm:
http://www.ina.gl/inatsisartuthome/sammensaetning-af-inatsisartut.aspx

wfdd:
Scraper for all terms at https://morph.io/wfdd/inatsisartut-scraper.

wfdd:
Also fixed #381 and the incorrect numbering of terms.

wfdd:
This can be closed now.

tmtmtmtm:
thanks @wfdd

PR #12032, opened 2016-06-10
Greenland from scraper

### LINKED ISSUE #381: Greenland: Party spillage into Person names
e.g. on http://everypolitician.org/greenland/inatsisartut/term-table/2005-11-15.html

screen shot 2015-07-16 at 13 58 57

wfdd:
Closes #381 and #5773.


wfdd:
There's 3 people it's not updated in https://everypolitician-viewe-pr-11221.herokuapp.com/greenland/inatsisartut/term-table/12.html - dunno what's going on there, it all looks fine in my data.

tmtmtmtm:
Woot! Thanks @wfdd 

Which three people do you mean on that page?

wfdd:
Mala Høy Kuko, Múte Bourup Egede and Nivi Olsen all have the old placeholder photo, so I assume it's not picked up my changes.

wfdd:
Never mind, it dawns on me the reason for that is the image cache is only updated periodically.

tmtmtmtm:
This is great, thanks! The switch from Landsting to Inatsisartut gives me a slight pause, as I'm not totally sure whether this was a simple renaming of the same body, or technically created a new body (in which case we might want to split it out into a separate directory). But that's something we can certainly return to later, as this is definitely an improvement over what we currently have.

I think you might want to look at URL encoding the image links, to protect against a lot of them having spaces in them (though it's also possible that we should do things like that at input so it gets applied everywhere…)

(BTW: the image proxy only reads from what's already in `master`, so is never quite right in the pull-review previews. There's presumably something fancy we could do there, but it's never really been a big enough of an issue to really pay it much attention).

wfdd:
The parliament was renamed a few weeks into a new term, which was when the self-rule act came into force.  Greenland has had home rule since 1979 and the 2009 act devolved even more powers to their government.  The new law made Greenlandic the official language and hence the renaming (see http://www.stm.dk/_a_2957.html).  In an [official publication](http://www.ina.gl/media/28274/Valg%20til%20Inatsisartut%20DA%20WEB.pdf) (pp. 31-33), they list all of the elections since 1979, followed by the parliament term they correspond to; the term beginning in 2009 was the 10th and the one beginning in 2013 the 11th.  In the same PDF, before 2009, the house leader is called the 'Landstingsformand' and thereafter 'Formand for Inatsisartut', though the parliament itself is called the Inatsisartut throughout the document.  Anyhoo, I think is probably enough evidence that they consider it to be a continuation of the pre-2009 parliament.

PR #12113, opened 2016-06-11
Wallis and Futuna (Territorial Assembly): Import gender and alternative names from official source

wfdd:


wfdd:
I thought I'd 'play' some of these lesser-known legislatures on Gender Balance, I started with WF, there weren't any photos, and now here we are...

tmtmtmtm:
Woohoo! 👍

PR #12496, opened 2016-06-17
Uganda (Parliament): Stop fuzzy matching OCD areas

chrismytton:
This was causing some bad data, e.g.:

```
Matched Area West Moyo County to Chua West County
Matched Area Oyam County North to Aruu County North
Matched Area Bukoto County South to Mukono County South
```

So instead of doing a fuzzy match we're now doing an exact match. This means we'll need to get an updated mapping file from Collin which maps to the correct area names.


chrismytton:
Closing as this code has drifted from `master` and I think we'll be doing https://github.com/everypolitician/everypolitician/issues/599 instead now.

PR #12795, opened 2016-06-22
Brazil (Deputies): add secondary official source

tmtmtmtm:
Add in the scraper that also picks up historic data, and cross-reconcile so that it can become the primary source.


tmtmtmtm:
None of the missing people are listed on https://pt.wikipedia.org/wiki/Lista_de_deputados_federais_do_Brasil_da_55.%C2%AA_legislatura, so I'm presuming this version is more correct.

PR #12906, opened 2016-06-24
Reconciler: don't consider skipped matches when listing duplicates

mhl:
If you skipped more than one match in the reconciler, they would be
reported a duplicate IDs for the UUID "null".  This commit filters out
such skipped votes before looking for duplicates in them.

Fixes everypolitician/everypolitician#367


chrismytton:
Looks good! 👍

tmtmtmtm:
@chrismytton is there a reason why this one wasn't merged?

chrismytton:
Nope, no reason!

PR #13414, opened 2016-07-05
Cabo Verde (Assembleia Nacional): refresh data

everypoliticianbot:
Automated data refresh for Cabo Verde - Assembleia Nacional

#### Output

```
  ⚥ data for 0; 0 added

Creating 2 term files
Creating names.csv
Persons matched to Wikidata: 0 ✓ | 81 ✘
Parties matched to Wikidata: 3 ✓ 

```


tmtmtmtm:
@wfdd this is from a clean run of the new scraper :(

Don't have a lot of time to go digging at the minute — anything obvious jump out at you?

wfdd:
Well, this is not good:

- I've my doubts the official list of MPs for the 8th term is accurate.  I can't find any reference to Walter Emanuel da Silva Évora online before the '16 race, for instance, and he's not in the old dataset either.
- The page IDs are specific to a term, so the new term's IDs need to be reconciled.
- It looks like the previous term's IDs have moved, so they too need to be reconciled - though if there's no gender data, I don't suppose it really matters.

wfdd:
Okay, so, the legislature selection doesn't stick when moving from page to page, which meeeeans the scraper's only scraped the first page of the 8th term's MPs correctly,

PR #13588, opened 2016-07-08
Ireland dáil twitter

davewhiteland:
Made PR for review (@tmtmtmtm) in anticipation of committing again with fixedup data -- I _think_ upstream scraper should be knocking out "bad" names like these, and instead fixing them to be canonical (i.e., vs what we've already got, right)? 

* MaryMitchellO'Connor (one word)
* ☎️ Josepha Madigan (crazy glyph)
* anything with Cllr or Councillor in the title

Putting this here to hang the discussion off it (and of my process on this style of PR in general).


tmtmtmtm:
our main options here seem to be:

1. use all the names 'as is' (on the basis that if this is what someone has called themselves on Twitter, then it _is_ a valid name for them, and it's only ever going to be listed as an alternative name anyway, not the primary one)
2. use none of the names by removing them from the SELECT, on the basis that we already _have_ names for these people anyway, so we don't need them.
3. Pick and mix which names actually add value, which essentially requires fixing them all up in the scraper.

3 is a lot more work, though I'm OK if you want to do it. If we want the simpler version I'd probably lean towards #2 in this specific case, as there don't seem very (m)any that are particularly useful to pick up. If there are ones we really like the idea of having as alternate names in EP, we could also add them as AKA labels in Wikidata…

davewhiteland:
@tmtmtmtm thanks for that -- yes as I went through these it became increasingly clear that twitter isn't really a useful source of what I'd expect a `name` to be in the Popolo, so I favour option 2. Most of them are really variations of the real name solely for the convenience of their use on twitter, i.e., they have no currency away from twitter.

I'll go ahead with 2.

tmtmtmtm:
superseded by #14399

PR #13594, opened 2016-07-08
Iran new term 10

davewhiteland:
Incoming data from ASL19's spreadsheet


data/Iran/Assembly/sources/manual/terms.csv
```diff
@@ -1,2 +1,3 @@
 id,name,start_date,end_date,source
 9,9th Assembly,2012-05-27,,
+10,10th Assembly,2016-05-28,,
```
tmtmtmtm: You need to add an end date for the 9th Assembly as well…

data/Iran/Assembly/sources/manual/terms.csv
```diff
@@ -1,3 +1,3 @@
 id,name,start_date,end_date,source
-9,9th Assembly,2012-05-27,,
+9,9th Assembly,2012-05-27,2016-05-26,
```
tmtmtmtm: This could probably do with a comment in the commit message for where you got the date from (IPU IIRC?)

davewhiteland:
Yes I will add a comment to the commit message about the source of the end-of-term data, and force push it amended.

Other data is good to go _except_ we're not parsing the incoming Iranian dates which I believe means end dates are being ignored, so the data includes a few members who are no longer extant.

I think this is good to go live with subsequent correction on those to follow but ymmv.

tmtmtmtm:
@davewhiteland I'm happy to put this live as is, as it's an improvement over what we currently have. But it would be good to have the ticket in place already for fixing up the dates, and linked from here. I'm also a little concerned that Travis is failing.

tmtmtmtm:
Ah, this branch is from before the addition of the `.travis.yml`, so it's failing on the push, but would pass on merge. I think we can safely ignore that, then, so the only issue is ensuring there's a ticket for fixing the dates. @davewhiteland once you create and link that, I'll merge this.

PR #13607, opened 2016-07-08
Add Colombian Senate

wfdd:


wfdd:
It might make sense to cast the `id`s to `INTEGER` to sort them numerically, if it matters at all.

tmtmtmtm:
Yay. Thanks for this!

> It might make sense to cast the ids to INTEGER to sort them numerically, if it matters at all.

I don't think it matters — a _consistent_ order is useful to eliminate diffs that are just things shuffling around, but what exactly that order is doesn't really matter.

wfdd:
Fair enough.  Let me know if anything needs changing.  By the way, I've noticed there's a general absence of upper houses - is that 'cause many aren't directly elected?

tmtmtmtm:
No — we largely just haven't gotten around to them yet, as we prioritised getting at least _something_ for every country over going particularly deep on each. (In practice it's also a little tricky because we still need a way to reconcile people across both houses, so they can have the same UUID everywhere, and the reconciler only works within a single legislature at the minute). There'll be more of a push on this again over the next couple of months, I would expect.

wfdd:
Oh right.  Thanks for explaining.

wfdd:
Now added Dave's Twitter scraper, which I'd completely forgotten about yesterday.  If someone wants to write one for Wikidata, I can add that too (or you can clone the branch and do whatever).

tmtmtmtm:
Thanks @wfdd — I'll merge this for now, and we can add Wikidata separately (feel free to write the scraper if you like — they're usually only about 3 lines long if you can find a good source — e.g. a Wikidata Category, or a page that lists them all in a table or somesuch)

PR #13660, opened 2016-07-09
Add Colombian Senate Wikidata

wfdd:


wfdd:
I've run it again and rebased:

```
Merging with sources/morph/twitter.csv
* 3 of 88 unmatched
	{:id=>"58982725", :name=>"Eugenio Prieto"}
	{:id=>"155964060", :name=>"#VoteBien"}
	{:id=>"1978999279", :name=>"Roosvelt Rodríguez"}
Merging with sources/morph/wikidata.csv
* 163 of 225 unmatched
	{:id=>"Q9025120", :name=>"Luis Guillermo Vélez"}
	{:id=>"Q116059", :name=>"Rodrigo Lara Bonilla"}
	{:id=>"Q2879824", :name=>"Cecilia López Montaño"}
	{:id=>"Q5885839", :name=>"Griselda Janeth Restrepo"}
	{:id=>"Q6103901", :name=>"Reginaldo Montes"}
	{:id=>"Q5951281", :name=>"Juan Martín Caicedo"}
	{:id=>"Q6299134", :name=>"Juan Camilo Restrepo Salazar"}
	{:id=>"Q3046401", :name=>"Luis Carlos Galán"}
	{:id=>"Q5703591", :name=>"Argelino Durán Quintero"}
	{:id=>"Q5552352", :name=>"Germán Zea Hernández"}
  ⚥ data for 0; 0 added

Creating 1 term file

Top identifiers:
  62 x wikidata
  5 x viaf
  4 x lcauth
  2 x gnd
  2 x isni

Creating names.csv
Persons matched to Wikidata: 62 ✓ | 39 ✘
  No wikidata: PRIETO RIVEROS JORGE (2a1772a0-d316-4957-b14d-1713198611a6)
  No wikidata: ANDRADE CASAMA LUIS EVELIS (37ce24df-0697-43dc-addc-988265194fb6)
  No wikidata: MARTÍN EMILIO MORALES DIZ (2741d2d6-6eb0-4107-90a5-a948ce85ca4e)
  No wikidata: ANTONIO JOSÉ CORREA JIMÉNEZ (1839ad5e-5b3f-48c7-8b0e-ba413ce458de)
  No wikidata: SEGUNDO SENEN AVENDAÑO NIÑO (13d00d8a-cc14-4060-8b38-b0e1c547524c)
  No wikidata: BARON NEIRA LEON RIGOBERTO (1f1f95b8-e7ba-430b-a601-3ffefee5ed3b)
  No wikidata: MARITZA MARTÍNEZ ARISTIZABAL (0c0fe943-db07-4371-ba13-a0456cc7d9c1)
  No wikidata: ERNESTO MACIAS TOVAR (23f217c2-45f4-483f-994f-d45b03edff21)
  No wikidata: YAMINA DEL CARMEN PESTANA ROJAS (53f690d0-ec9f-4d75-92c5-bf7cd5de41c6)
  No wikidata: JULIO MIGUEL GUERRA SOTTO (0fb6c5d7-16d4-43e3-9ba7-ecb20f8397b3)
Parties matched to Wikidata: 0 ✓ | 20 ✘
```

tmtmtmtm:
Thanks @wfdd 

Was the wikidata query the best set of results for these? It's quite unusual for Wikidata to be well enough maintained in most countries for that to be good enough to rely on. It looks like the Spanish 'Categoría:Senadores de Colombia 2014-2018' might be useful, either as a replacement, or combined with with the WDQ version.

wfdd:
Good enough in what respect?  Numerically?  The Spanish category has 98 members and I've only been able to match 62 via the item query, so the category looks like a better option.  OTOH, I think it'd make more sense to tag the remaining senators on Wikidata than to scrape Wikipedia directly 'cause then it'd also be doing our part in getting Wikidata up to speed.

tmtmtmtm:
Yes — I just picked a couple of people from the list of unmatched ones and tried to see if I could find them some other way, and ended up at that Category.

Adding them Wikidata is definitely a good thing, but it's not just about what's _currently_ in Wikidata, it's also about future proofing — when a new Senator starts, at this stage it's considerably more likely that they get added to the Category on Wikipedia than they get the correct P39 set on Wikidata…

wfdd:
It'd also be helpful if the reconciler stripped diacritics.  I'll look into it...

tmtmtmtm:
https://github.com/everypolitician/everypolitician/issues/404 is about that (my presumption is that this is only really worth doing at the display stage, as doing it at when scoring matches seems a little more problematic, but open to persuasion if you were considering doing it earlier)

tmtmtmtm:
> There's 15 it didn't fuzz 'cause they had no English names 

That's one of the disadvantages of looking up directly by ID. If you tweak the scraper to something more like this (untested), it should use the Spanish names by default instead (it chooses names based on the order of the keys in the `names` hash, which was a sensible way to do things when the `names` hash was the _only_ option. That doesn't work so well when looking up by `id` instead, but I think this is the first time we've hit that problem)

```ruby
ids =  EveryPolitician::Wikidata.wdq('claim[39:19254253]')
names = WikiData::Category.new('Categoría:Senadores de Colombia 2014-2018', 'es').member_titles

EveryPolitician::Wikidata.scrape_wikidata(ids: ids, names: { es: names })
```

wfdd:
Alright, that's kind of a convoluted way to prioritise `es`.  Fixed in wfdd/colombia-senado-wikidata-scraper@e3ca101 and I'll update the data when I get home.

tmtmtmtm:
> that's kind of a convoluted way to prioritise es

If you weren't also fetching by ID (which is a relatively recent extension) the script would just have been:

```ruby
names = WikiData::Category.new('Categoría:Senadores de Colombia 2014-2018', 'es').member_titles
EveryPolitician::Wikidata.scrape_wikidata(names: { es: names })
```

This would then use the Spanish labels by default, which is almost always what we want.


Or, if combining data from multiple sources (e.g. Wikipedia pages in multiple languages — for example with [Abkhazia](https://github.com/tmtmtmtm/abkhazia-peoples-assembly-wikidata/blob/master/scraper.rb)):

```ruby
scrape_wikidata(names: { en: en_names | en_names2, ab: ab_names, ru: ru_names })
```

Until very recently that was the _only_ way these could be created, and so we chose to prioritise generating the `name` field by the order of those keys. (i.e. if we wanted the Abkhaz labels by default, we could switch that around to:

```ruby
scrape_wikidata(names: { ab: ab_names, en: en_names | en_names2, ru: ru_names })
```

This does The Right Thing™ 99% of the time, so was worth the simplicity IMO. Adding the ability to look up by ID as well as (or instead of) names has definitely made this a little more awkward in those cases, though, so I'm definitely open to other more explicit approaches — for example an extra argument to `scrape_wikidata` that gives an ordered list of language codes to try when constructing the `name` field from the Wikidata labels.

wfdd:
> for example an extra argument to scrape_wikidata that gives an ordered list of language codes to try when constructing the name field from the Wikidata labels

That already exists, `lang` (though it should really be `langs` since it takes an array of language codes as input), which overrides the `names` keys.  It shouldn't be difficult to decouple the two.

wfdd:
Alright, 95 matched to Wikidata.  The remaining three in that category we've not paired with anybody are: [Ana Mercedes Gómez](https://es.wikipedia.org/wiki/Ana_Mercedes_Gómez), [Eugenio Prieto Soto](https://es.wikipedia.org/wiki/Eugenio_Prieto_Soto) and [Roosvelt Rodríguez](https://es.wikipedia.org/wiki/Roosvelt_Rodr%C3%ADguez).  The first two have resigned.  I've no idea why Roosvelt Rodríguez doesn't have a page on the Senate website, but here we are.  I'll rebased if we're ready to merge.

tmtmtmtm:
Looks great to me!

wfdd:
Rebased.

tmtmtmtm:
:+1:

Thanks @wfdd

wfdd:
> everypolitician/everypolitician#404 is about that (my presumption is that this is only really worth doing at the display stage, as doing it at when scoring matches seems a little more problematic, but open to persuasion if you were considering doing it earlier)

I'd suggest implementing it as an option at the scoring stage.  Capitalised names tend to be accented inconsistently, and if the input is an abbreviated name and the name already in the database is a Spanish full name, covariance will be high and one wrong letter could bump the right match off the list of suggested matches.

PR #13834, opened 2016-07-12
WIP: Group wikidata refactor

chrismytton:
Extract `GroupPopolo` class which can be used to cleaning and augmenting the groups/parties int he popolo.

I'm not 100% sure about the name of this class or the name of the method, so this pull request is a bit of a straw man.


tmtmtmtm:
@chrismytton : what's the first change here? Is that something that's broken already in the baseline? If so that should probably be a separate PR

chrismytton:
> what's the first change here? Is that something that's broken already in the baseline? If so that should probably be a separate PR

Yep, good shout. Have extracted that change in https://github.com/everypolitician/everypolitician-data/pull/13949.

tmtmtmtm:
I'm not sure I know what the names should be either, so maybe best to just have a big obvious comment that says so, so we think about it again later, and see if it becomes more obvious as we take the next steps?

PR #13949, opened 2016-07-14
Fix broken reconciler test and setup Travis CI

chrismytton:
Fix test that was failing on `master`.

I've also added a `.travis.yml` and a default rake task so the tests can run on Travis for each PR.


tmtmtmtm:
👍

PR #13993, opened 2016-07-15
Add an Enumerable#portable_sort_by method

chrismytton:
This is a workaround for [sort_by working differently between OS X and
Linux](https://bugs.ruby-lang.org/issues/11379).

### DIFF
--- a/rake_build/generate_final_csvs.rb
+++ b/rake_build/generate_final_csvs.rb
@@ -60,7 +60,7 @@ namespace :term_csvs do
     terms.each do |t, rs|
       filename = "term-#{t}.csv"
       header = rs.first.keys.to_csv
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
There's a separate commit coming for actually using this? Or you want me to take over that bit?

chrismytton:
I've just pushed a commit which fixes the UK/Commons build. I'm guessing other legislatures might produce diffs for other fields, so we should probably replace all the `sort_by!` calls in `rakefile_common.rb` with `portable_sort_by`.

@tmtmtmtm If you could take over and change the other bits of `rakefile_common.rb` that'd be much appreciated 👍

tmtmtmtm:
This now fixes all the sort-related build inconsistencies in https://github.com/everypolitician/everypolitician/issues/435

chrismytton:
Looks good 👍

Issue #13831, opened 2016-07-12
Nauru new term


wfdd:
A parliamentary election was held in Nauru on 9/7.

https://news.google.com/news/search?tbm=nws&q=nauru%20election%20results

tmtmtmtm:
Thanks @wfdd — did you happen to see actual usable (ie scrapable) data in any of those pages? Most of the ones I've seen just happen to mention some key facts in a wall of text, rather than (say) a handy table of all the results.

(NB not sure if you noticed, but the EP data now includes general election info for most countries, in large part so that we can actually get warnings of which countries have had more recent elections without us having up to date data: https://github.com/everypolitician/everypolitician-data/blob/5c76f578c34a243950aa62f744a11fff376b6648/data/Nauru/Parliament/ep-popolo-v1.0.json#L4087-L4093)

wfdd:
There's a list of elected candidates on [Wikipedia](https://en.wikipedia.org/wiki/Nauruan_parliamentary_election%2C_2016).  They don't cite any sources.

> NB not sure if you noticed ...

Oh I see.  Are those warnings automated/public?

wfdd:
http://www.naurugov.nr/government-information-office/media-release/elected-members-for-nauru-elections-2016-(1).aspx

tmtmtmtm:
Excellent, thanks

> Are those warnings automated/public?

Not yet. But the [per-legislature stats files](https://github.com/everypolitician/everypolitician-data/blob/master/data/Nauru/Parliament/unstable/stats.json) includes the start date of the last term, and the date of the last election (even if that's in the future, as it's useful to be forewarned of upcoming elections). At the moment we just pull those into a local spreadsheet with automated colour coding, as a "simplest thing" first step. Eventually we want to have a nice on-line dashboard, but that's not super-high priority atm.

PR #14054, opened 2016-07-16
Nauru: Add new Term

tmtmtmtm:
Source: http://www.naurugov.nr/government-information-office/ (thanks @wfdd)

Scraper: https://morph.io/everypolitician-scrapers/nauru-2016-results

Closes https://github.com/everypolitician/everypolitician-data/issues/13831


wfdd:
Are those numbers actually votes?  Even with preferential voting (what do they use, STV?), I don't see how there can be more votes than people.

wfdd:
Err, what I meant to say is each candidate couldn't have received more votes than there are people.

wfdd:
Oh the dot you're stripping off on line 56 of the scraper is the decimal mark - I suppose you thought it was the thousands delimiter?  According to the all-knowing Wikipedia, they divide the vote by candidate's rank on the ballot.

tmtmtmtm:
Thanks, fixed (though we don't actually use it anywhere)

PR #14261, opened 2016-07-20
Add twitter list scrapings to Australian Senate

ondenman:
Add twitter data from https://twitter.com/AuSenate/lists/senators
via scraper at https://morph.io/everypolitician-scrapers/twitter-australia-senate


tmtmtmtm:
@ondenman it's good practice to include a description of what the PR is — in this case it's usually enough to link to the scraper and the list (similar to https://github.com/everypolitician/everypolitician-data/pull/14242)

tmtmtmtm:
Probably worth noting that https://twitter.com/nthaustralia is clearly also Ian MacDonald, but appears to be obsoleted by https://twitter.com/senatormac but this comment will do for that, so as soon as there's a better description of the PR up to, this is 👍  from me!

data/Australia/Senate/sources/instructions.json
```diff
@@ -71,6 +71,21 @@
         "from": "wikidata-raw",
         "source": "reconciliation/wikidata.csv"
       }
+    },
+    {
+      "file": "morph/twitter.csv",
+      "create": {
+        "from": "morph",
+        "scraper": "everypolitician-scrapers/twitter-australia-senate",
+        "query": "SELECT * FROM data"
```
tmtmtmtm: To avoid false positive diffs later, where the only change is the order that lines come out the database, it's good practice to always add an explicit `ORDER BY` here: e.g. https://github.com/everypolitician/everypolitician-data/blob/ad7e0b9d36826a387f957e14c4b3aa3aa5b202c4/data/Scotland/Parliament/sources/instructions.json#L32

PR #14263, opened 2016-07-20
Add twitter list scrapings to Welsh Assembly

ondenman:
Add twitter data from https://twitter.com/AssemblyWales/lists/assembly-members
via scraper at https://morph.io/everypolitician-scrapers/welsh-national-assembly-twitter-list


tmtmtmtm:
@ondenman is this still in progress? Or is this ready for review? (Pending adding a proper description etc, of course)

tmtmtmtm:
The one duplicate in this is that Michelle Brown is listed as both [WalesUKIP](https://twitter.com/WalesUKIP)` and [@UKIPNorthWales](https://twitter.com/UKIPNorthWales)

The former looks like it was (or was to be, depending on whether it was never operational, or whether the tweets were deleted) a campaign-related account, whereas I think the latter is the only one we want.

PR #14266, opened 2016-07-20
Reconciler: Ensure `;` separated fields are rendered correctly

chrismytton:
Try to split the fields on the `;` and display all parts separately, rather than as one.

screen shot 2016-07-20 at 16 58 59

- [ ] This code has quite a bit of duplication, so could do with a refactor to tidy up some of that and move the logic out of the HTML and into the JavaScript

Fixes https://github.com/everypolitician/everypolitician/issues/469

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -79,7 +79,9 @@
                   <dt title="{{ field }}">{{ field }}:</dt>
                   <dd>
                   {% if (field == 'image') { %}
-                    <img src="{{ person[field] }}" width="150">
+                    {% person[field].split(';').forEach(function(img) { %}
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
@chrismytton I'm going to go ahead and merge this, as it's a much needed improvement, and because I fear that your desire to remove the duplication here is actually premature abstraction. The LHS and RHS are actually conceptually quite different things, that happen _look_ very similar at the moment, because we're not doing very much with them, but abstracting out commonalities now carries a very high risk of making it much harder later to actually evolve each separately.

tmtmtmtm:
I do agree that more of the logic should live elsewhere though, so feel free to open a new ticket for that!

PR #14269, opened 2016-07-20
update Burkina Faso term dates to be more accurate

### LINKED ISSUE #13994: Burkina Faso: missing end-of-term dates
Looks like we need to add end-of-term dates to all the National Assembly terms (currently none of them have them)

davewhiteland:
Specifically this is providing end dates that were previously both fuzzy (only years) and missing (missing `end_date` heading in the CSV, whoops)

Closes #13994


tmtmtmtm:
👍  — thanks @davewhiteland

PR #14273, opened 2016-07-20
Use finer-grained git commit object names in file URLs

### LINKED ISSUE #451: Bulgaria
Here's [my own scraper](https://github.com/briatte/bgparl/blob/master/data.r).

Links to MP pages are of the form

http://www.parliament.bg/bg/MP/222

The index is at

http://www.parliament.bg/bg/MP/

mhl:
The `countries.json` file includes the URLs of files via
cdn.rawgit.com, and those URLs include the commit's object name (also
known as its hash or SHA-1) so that a known version of every file is
referred to in each version of countries.json.

The commit object names used in these URLs (for both the Popolo JSON
and term CSV files) would always be the same for all files in a
particular country, since the commit and last modification time would be
found from a command like:

    git --no-pager log --format='%h|%at' -1 data/Australia/

... and the results would be used for all files for that country.

This meant that if any file for that country was updated, then all files
for that country would have a new commit object name in their URL. This
was bad, because a consumer of that data may be using changes in these
URLs to detect if there is new data to process, and this would mean
they'd have to process more data than necessary.

This commit changes that: now when rebuilding the countries.json file a
mapping is first found from each filename under data/ to the most recent
(based on committer date) non-merge commit that changed that file. This
mapping is then used to find the commit by which a file should be
referred to on a per file basis.  This fixes everypolitician/everypolitician#451.

Note that both this and the previous version of the code are incorrect,
strictly speaking. There is no guarantee that the commit found by
either method has the same version of the file as in HEAD. For example,
both methods ignore merge commits, and it's possible that a merge commit
has a different version of the file than the most recent non-merge
commit. Also, the order by committer date can be completely different
from topological order. Because of the workflow by which merges are done
in this repository, however, this isn't likely to cause a problem in
practice, but a more correct way to do this would be:

  - Find the object name of the file's blob in HEAD

  - Find the earliest commit (by whatever ordering) which has that
    blob's object name at that path.

That's rather more awkward to implement, however, so this version should
do for the moment, and, as I said, it's no *more* incorrect than the
previous version.

In terms of performance, this is only a few seconds slower on my laptop
than the previous version; that cost is is dwarfed by the time taken to
reclone the repository each time - see
https://github.com/everypolitician/everypolitician/issues/359

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -4,6 +4,7 @@ require 'pathname'
 require 'pry'
 require 'tmpdir'
 require 'json'
+require_relative 'lib/git'
 
 ISO = IsoCountryCodes.for_select
 
@@ -61,13 +62,18 @@ task 'countries.json' do
   end
 
   data, _ = json_from('countries.json') rescue {}
-  
+  # If we know we'll need data for every country directory anyway,
+  # it's much faster to pass the single directory 'data' than a list
+  # of every country directory:
+  commit_metadata = file_to_commit_metadata(
+    to_build == 'data' ? ['data'] : countries.values.flatten
+  )
+
   countries.each do |c, hs|
     meta_file = hs.first + '/../meta.json'
     meta = File.exist?(meta_file) ? JSON.load(File.open meta_file) : {}
     name = meta['name'] || c.tr('_', ' ')
     slug = c.tr('_', '-')
-
     country = {
       name: name,
       # Deprecated — will be removed soon!
@@ -79,9 +85,7 @@ task 'countries.json' do
         name_file = h + '/names.csv'
         remote_source = 'https://cdn.rawgit.com/everypolitician/everypolitician-data/%s/%s'
         popolo, statement_count = json_from(json_file)
-
-        cmd = "git --no-pager log --format='%h|%at' -1 #{h}"
-        (sha, lastmod) = `#{cmd}`.chomp.split('|')
+        sha, lastmod = commit_metadata[json_file].values_at :sha, :timestamp
         lname = name_from(popolo)
         lslug = h.split('/').last.tr('_', '-')
         {
@@ -94,7 +98,10 @@ task 'countries.json' do
           lastmod: lastmod,
           person_count: popolo[:persons].size,
           sha: sha,
-          legislative_periods: terms_from(popolo, h).each { |t| t[:csv_url] = remote_source % [sha, t[:csv]] },
+          legislative_periods: terms_from(popolo, h).each { |t|
+            term_csv_sha = commit_metadata[t[:csv]][:sha]
+            t[:csv_url] = remote_source % [term_csv_sha, t[:csv]]
+          },
           statement_count: statement_count,
         }
       }
--- a/lib/git.rb
+++ b/lib/git.rb
@@ -1,0 +1,35 @@
+# A helpful monkey-patch to MatchData to return named matches as a
+# hash that maps from symbolized match names to the matched text.
+class MatchData
+  def to_hash
+    Hash[ (self.names.map(&:to_sym)).zip(self.captures) ]
+  end
+end
+
+def file_to_commit_metadata(directories)
+  # There's about 2MB of data returned by this command, so rather than
+  # using backticks to turning it into a string and then splitting it,
+  # parse the output line-by-line. As a bonus this doesn't
+  # unnecessarily invoke a shell.
+  command = [
+    'git', '--no-pager', 'log', '--name-only', '--format=%h|%at',
+    '--', *directories
+  ]
+  last_commit = {}
+  commit_details = nil
+  IO.popen(command) do |f|
+    f.each_line do |line|
+      # Each commit is introduced with the abbreviated object name
+      # of the commit and author date timestamp, separated by '|'.
+      # Then there's a blank line, then one filename per line.
+      line.strip!
+      commit_match = line.match /^(?<sha>[a-f\d]+)\|(?<timestamp>\d+)$/
+      if commit_match
+        commit_details = commit_match.to_hash
+      elsif not line.empty?
+        last_commit[line] ||= commit_details
+      end
+    end
+  end
+  last_commit
+end


tmtmtmtm:
Thanks @mhl 

I'll look more closely at the code later, but as a quick note, you're closing Issue 451 in the wrong repo with this at the moment.

tmtmtmtm:
I've also assigned @chrismytton, as it would be good to get his feedback on this as well.

Rakefile.rb
```diff
@@ -48,6 +48,36 @@ def name_from(json)
...
+      # In the parent process after forking, parse the output:
+      f.each_line { |line|
+        commit_match = line.match /^([a-f\d]+)\|(\d+)$/
+        if commit_match
+          sha, timestamp = commit_match.captures
+        else
+          filename = line.strip
+          unless result.has_key? filename
+            result[filename] = {sha: sha, timestamp: timestamp}
+          end
```
tmtmtmtm: I think this would be clearer, and more Rubyish as simply
```ruby
  result[filename] ||= { sha: sha, timestamp: timestamp } 
```
I'd probably drop the temporary `filename` variable too:
```ruby
else
  result[line.strip] ||= { sha: sha, timestamp: timestamp }
end
```
but that's less clear-cut, as at the minute it helps document the input format a little. I suspect, though, it might be better still to make it clearer at a higher level what that format is, as that's the part I'm finding it hard to grok initially.

Rakefile.rb
```diff
@@ -48,6 +48,36 @@ def name_from(json)
...
 end
 
+def file_to_commit_metadata
+  # There's about 2MB of data returned by this command, so rather than
+  # using backticks to turning it into a string and then splitting it,
+  # parse the output line-by-line. As a bonus this doesn't
+  # unnecessarily invoke a shell.
+  command = ['git', '--no-pager', 'log', '--name-only',  '--format=%h|%at', '--', 'data']
+  result = {}
+  sha, timestamp = nil
```
tmtmtmtm: This does _work_, but it feels odd to me because it looks like you're expecting the assignment to apply to both elements on the left hand side, with the implication that after `foo, bar = 1`, both `foo` and `bar` would each be set to `1`. 

`sha, timestamp = nil, nil` would be more correct, I think, but as you never actually do anything with those variable separately, I suspect a cleaner approach would be to keep them paired as a `commit_details` hash instead:

```
      f.each_line do |line|
        commit_match = line.match /^([a-f\d]+)\|(\d+)$/
        if commit_match
          commit_details = { sha: commit_match[1], timestamp: commit_match[2] }
        else
          result[line.strip] ||= commit_details
        end
      end
```

Though that might be clearer still with named captures:
```
      f.each_line do |line|
        commit_match = line.match /^(?<sha>[a-f\d]+)\|(?<timestamp>\d+)$/
        if commit_match
          commit_details = { sha: commit_match[:sha], timestamp: commit_match[:timestamp] }
        else
          result[line.strip] ||= commit_details
        end
      end
```

(NB: For maximum readability I'd be tempted to actually monkey patch MatchData with: 
```ruby
def to_hash
  Hash[ (self.names.map { |n| n.to_sym }).zip( self.captures ) ]
end
```

But that might be a step too far…)

Rakefile.rb
```diff
@@ -48,6 +48,36 @@ def name_from(json)
...
+            result[filename] = {sha: sha, timestamp: timestamp}
+          end
+        end
+      }
+    else
+      # In the child process after working, exec the command:
+      exec *command
+    end
+  }
+  result
```
tmtmtmtm: I think I'd be tempted to rename `result` to something that more adequately reflects _what_ it's actually storing, rather than just the fact that it's a holder for the return value. 

Maybe something like `last_commit`, so that the assignment reads:
```ruby
   last_commit[filename] ||= { … }
```
Something like that is likely to be a bit easier to grasp on an initial scan of the code before you've loaded the entire context of what this doing.

Rakefile.rb
```diff
@@ -61,13 +91,13 @@ def name_from(json)
   end
 
   data, _ = json_from('countries.json') rescue {}
+  commit_metadata = file_to_commit_metadata
```
tmtmtmtm: In day to day usage (i.e. not the full build that the bot runs), we tend to run this with the `EP_COUNTRY_REFRESH` env variable set, to only update a single country.{1}

At the minute this isn't making any use of that. I think we want to have `file_to_commit_metadata` take a parameter for what directories to examine, and then pass those into the `git log`:

```ruby
commit_metadata = file_to_commit_metadata(countries.values.flatten)
…
def file_to_commit_metadata(dirs)
  command = ['git', '--no-pager', 'log', '--name-only',  '--format=%h|%at', '--', *dirs]   
```

That makes this run in about half the time on my laptop. 

---
{1} technically it could be more than one country as this is just a simple `include?` check, but the standard approach is something like `EP_COUNTRY_REFRESH=Estonia be rake countries.json`

Rakefile.rb
```diff
@@ -48,6 +48,36 @@ def name_from(json)
...
 
+def file_to_commit_metadata
+  # There's about 2MB of data returned by this command, so rather than
+  # using backticks to turning it into a string and then splitting it,
+  # parse the output line-by-line. As a bonus this doesn't
+  # unnecessarily invoke a shell.
+  command = ['git', '--no-pager', 'log', '--name-only',  '--format=%h|%at', '--', 'data']
+  result = {}
+  sha, timestamp = nil
+  Kernel.open('|-', 'r') { |f|
```
tmtmtmtm: This is a level of Ruby that I'm not entirely familiar with, but what's the advantage of dropping right down to this, rather than, say,
```ruby
    IO.popen(command) do |pipe|
      pipe.each_line do |line|
```
which seems to do the same thing, but without needing the extra level of `if f` indirection?

tmtmtmtm:
I've left a bunch of comments in-place. Most of them are fairly cosmetic, to hopefully make this easier to read and maintain later. However, the one about only fetching data on specified countries is quite important, I think.

Conceptually this also seems sufficiently complex, yet self-contained enough, to make it a perfect candidate for hiding^Wencapsulating into a separate class in the `lib` directory, so as to not distract/trip-up someone casually browsing the Rakefile later for other reasons. You shouldn't feel that it's necessary to do that in this PR, though; I'm happy enough to pull it out like that separately if you want to get on with other things.

Rakefile.rb
mhl: I've dealt with those comments in 5fa50f26277f42b19a hopefully - to mitigate it being less clear what `line` means, I added a comment about the format of the data. Also I noticed that the empty string is being uselessly set for each blank line, so that commit also includes a change to not process blank llines.

Rakefile.rb
mhl: Oops, yes. I like the monkey-patch (apart from my usual reservation about never being clear how to be sure if some other code is patching it differently :)), so I've added that - thanks for the suggestion. The fixup commit is: 010cf76b979aa29c25d785

Rakefile.rb
mhl: Done in 3596b91cd965f5

Rakefile.rb
mhl: The only issue with that change that I can see is that if you're rebuilding everything, the `git log` is about 4 times slower with each country directory listed individually rather than just the `data` directory, so I've changed it to just pass the `data` directory if `to_build == 'data'`. That change is in 3dd27b21106a08.

Rakefile.rb
mhl: I think you're right, there's no advantage to the more explicitly fork-and-exec code here, so I've switched it to use `IO.popen`; indeed it's much easier to read. Fixed in: 7afd601f861d8b4634

mhl:
I did the move to `lib` in b31ef85aba7cef8e7129 - does that look right?

I've fixed the issue reference in the PR message, and I'll fix it in the commit message when rebasing if that all looks OK.

tmtmtmtm:
@mhl Tiny nit on formatting, but all Looks Good To Me! 

If you let me know when it's ready to go live I'll do the actual merge as we probably want to prevent the bot updating everything in one go, and let the change ripple out slowly (else it'll cause awkward conflicts with lots of the other open PRs).

mhl:
@tmtmtmtm OK, I've rebased it onto current master, squashin in all those fixes including the additional formatting one you mentioned.

tmtmtmtm:
Now merged — thanks @mhl

PR #14314, opened 2016-07-21
Twitter list belgium representitives

ondenman:
Source: https://twitter.com/vlaparl/lists/vlaamse-regering
Scraper: https://morph.io/everypolitician-scrapers/twitter-list-belgium-vlaamse-regering


tmtmtmtm:
@ondenman — the list being scraped here is actually the Flemish Parliament, not the Belgian one, which is why you're only getting a single match…

tmtmtmtm:
If we were getting more data from it, it might be vaguely plausible, but for a single entry it's not worth it.

To make sure we don't lose the one piece of information it _did_ give it, though, I've added Annemie Turtelboom's twitter handle to her [Wikidata page](https://www.wikidata.org/wiki/Q454893), so we'll pick that up on next refresh.

PR #14316, opened 2016-07-21
Reconciler: Always show twitter links, and make them clicky

tmtmtmtm:
When generating cards in the Reconciler, we previously only included fields
that were common between the incoming card and at least one of the
comparisons. However, there are some fields we should _always_ display. Let's start with images and twitter handles, and make the twitter links clickable.

screen shot 2016-07-21 at 13 21 00

Closes https://github.com/everypolitician/everypolitician/issues/468

### DIFF
--- a/templates/reconciliation.html.erb
+++ b/templates/reconciliation.html.erb
@@ -84,11 +84,7 @@
                     {% }) %}
                   {% } else if (field == 'twitter') { %}
                     {% person[field].split(';').forEach(function(twitter) { %}
-                      {% if (twitter.startsWith('http')) { %}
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
@davewhiteland does this suffice for https://github.com/everypolitician/everypolitician/issues/468? (Feel free to open another ticket for the deluxe version, but I suspect that should be as part of the  twitter-list scrapers, whilst we're already hitting the API anyway, rather than any of the Reconciler's business)

@chrismytton Review welcomed!

chrismytton:
LGTM! 👍

davewhiteland:
@tmtmtmtm subject to me actually using it, this looks like it certainly suffices 👍

PR #14351, opened 2016-07-22
twitter-list-new-zealand-parliament-new-mps-as-at-20-10-2014

ondenman:
Add twitter data from https://twitter.com/NZParliament/lists/new-mps-as-at-20-10-2014
via scraper at https://morph.io/everypolitician-scrapers/twitter-list-new-zealand-parliament


tmtmtmtm:
I've rebased this over master, but there are still conflicts, so I'm assuming something is clashing with changes to NZ data that happened in the last couple of days.

This is where you'll the value of well separated commits, as you should now be able to simply discard the final commit, rebuild the data, and recommit the new version of all the output files.

tmtmtmtm:
@ondenman looks like you managed to delete lots of files in this one rather than rebuilding them

tmtmtmtm:
👍

PR #14359, opened 2016-07-22
twitter list Estonia

ondenman:
Add twitter data from https://twitter.com/electionista/lists/ee
via scraper at https://morph.io/everypolitician-scrapers/twitter-list-estonia


tmtmtmtm:
I've rebased this one over `master`

tmtmtmtm:
Looks good, except you missed this one:

screen shot 2016-07-22 at 11 55 11

I'm guessing that because you were operating off an old `master` you didn't have any of the Reconciler improvements from the last few days, that would have made this more obvious.

tmtmtmtm:
@ondenman after a review, you should make new commits with your changes, rather than squashing them onto pre-review code. That way the reviewer can look at what has changed since before, rather than having to go back and start over (in case, for example you've accidentally removed or changed other things too).

If they need to be squashed down again, that should be done at the end, when we're merging.

tmtmtmtm:
Also, when you force-pushed this back, you reset it to being based on Wednesday's master. I've rebased it again, and generated a fresh `countries.json` for it, so it's all good to go!

PR #14489, opened 2016-07-25
Turkey (Assembly): Add Twitter list scraper to sources

octopusinvitro:
Scraper here: https://morph.io/everypolitician-scrapers/twitter-list-turkish-politicians


tmtmtmtm:
I'm rather suspicious when I see twitter handles being added to members of significantly earlier terms. Mapping 'aliyoruger' to the Mehmet Ali Yörüker in the 5th-8th Assembly, for example, seems unlikely, as he [died in 1976](https://tr.wikipedia.org/wiki/Mehmet_Ali_Y%C3%B6r%C3%BCker).

For legislatures like this, where we have a _lot_ of historic data, but often not very much information about people (either at all, or visible in the reconciler), and where the incoming data includes many people who aren't members of the legislature (thus making it harder to even rely on two people with the same name being very likely to be the same person), I think we need to be considerably more explicit about confidence, both for the person immediately reviewing, and for people later trying to work out why we've matched certain people up.

So I think I'd do a first pass that checked in the ones which matched by twitter handle; then a second pass for people where I was very confident they're the same person; then, for ones that actually required research to decide, e.g. between multiple members with the same name, I'd be inclined to even make individual commits with explanations in the commit messages. 

Twitter handles aren't quite as important to get correct as some other fields, but in general false negatives are better than false positives (i.e. it's better to skip matches where we're not confident than to have false information). We also don't absolutely need to match everyone in one go — it's perfectly acceptable to match only people we're pretty sure of, and open another ticket to come back later to the others. The goal here is to deliver value quickly, not to achieve perfection. One good way to do that is to pick as much low hanging fruit as possible first, so we can get at least _some_ new data live, and then double back to the harder cases.

PR #14490, opened 2016-07-25
459 remove mailto prefix from email values

ondenman:
Some email address are prefixed with mailto:.
This fixes the problem by removing mailto: occurrences when data is built.

Closes https://github.com/everypolitician/everypolitician/issues/459


rake_build/combine_sources.rb
```diff
@@ -371,7 +371,10 @@ def combine_sources
     # Then write it all out
     CSV.open("sources/merged.csv", "w") do |out|
       out << all_headers
-      merged_rows.each { |r| out << all_headers.map { |header| r[header.to_sym] } }
+      merged_rows.each { |r|
+        r[:email].gsub!("mailto:","")
```
tmtmtmtm: I think this block should kept clean for simply doing the writing, rather than throwing extra logic in here. 

We really have three main choices for where we do this substitution:

1. When generating the `merged.csv` file (this file)
2. After we get JSON back from `csv-to-popolo` (`generate_ep_popolo.rb`)
3. Inside `csv-to-popolo`

On the basis that I think this substitution should really happen for _anyone_ who's generating Popolo — not just within EveryPolitician, I'd go with option 3.

ondenman:
Closing as fixed in: https://github.com/tmtmtmtm/csv_to_popolo/pull/102

PR #14399, opened 2016-07-23
Ireland (Dáil): refresh data from new Twitter list (donie/32d)

davewhiteland:
Very clean reconciliation: 1:1 mapping on everything from the list, with only one intervention/investigation required: Pat Gallaghar aka "Pat the Cope" (http://patthecope.com/) (UUID `d73f6aa3-349c-41c9-bb1b-529420e795a0`)

Note I'm _not_ importing `name` from this twitter list to avoid some unhelpful ones (e.g., "☎️ Josepha Madigan"), but the key one ("Pat the Cope") is already in the Popolo as an "other name" (thanks Wikidata!)

PR #14491, opened 2016-07-25
Add twitter names to Italy

ondenman:
Add twitter data from https://twitter.com/Montecitorio/lists/deputati
  via scraper at https://morph.io/everypolitician-scrapers/italy-twitter-list


tmtmtmtm:
The reconciliation itself looks great here, but the names we're picking up are generally of quite low quality. I think this is one where we want to do the name aliasing trick that @davewhiteland did in https://github.com/everypolitician/everypolitician-data/pull/14399

tmtmtmtm:
If we actually want to "rescue" any of these names, I'd suggest adding them as akas to their Wikidata page.

tmtmtmtm:
@ondenman Looks like the names are still all here…?

ondenman:
Closing this PR as I've obviously gone wrong somewhere. Easier to start again.

PR #14567, opened 2016-07-27
Twitter list South Korea

ondenman:
Add twitter data from https://twitter.com/teampopong/lists/national-assembly-19
  via scraper at https://morph.io/everypolitician-scrapers/twitter_list_South_Korea


tmtmtmtm:
@ondenman You only managed to reconcile 9 people here? 200 of them reconcile clearly even just against the _existing_ twitter data we have, even before we get as far as reconciling by name…

tmtmtmtm:
Looks great to me!

(Though I think it's probably worth noting that psh365 and ljy2111 both map to legislators with exactly the same name in Korean, but where it's not clear that they're actually the same person…)

PR #14568, opened 2016-07-27
Reconciler: standardise Twitter handles when comparing

tmtmtmtm:
If we’re comparing twitter handles, it’s _much_ better if they’re actually
in the same format, rather than comparing "tmtm" with "https://www.twitter.com/tmtm"

screen shot 2016-07-27 at 12 29 09

Closes https://github.com/everypolitician/everypolitician/issues/476


chrismytton:
Looks good! 👍

PR #14667, opened 2016-07-28
459 update csv to popolo version number

ondenman:
Updated cvs-to-popolo version number in gem file. New version removes `mailto:` from front of email addresses.

Closes https://github.com/everypolitician/everypolitician/issues/459


tmtmtmtm:
Looks good, but it would be a good idea to include one of the legislature here where we're expecting this to take effect, to show that it actually _does_.

tmtmtmtm:
👍

PR #14881, opened 2016-08-01
Twitter list italy 010816

ondenman:
Add twitter data from https://twitter.com/Montecitorio/lists/deputati
via scraper at https://morph.io/everypolitician-scrapers/italy-twitter-list

Skipped accounts:

Daniel Alfreider
@DanielAlfreider -- inactive account, no tweets

GERO GRASSI
@GRASSIGERO -- inactive account, one tweet in 2012


tmtmtmtm:
@ondenman I'm afraid I'm getting quite a lot of differences from you when I reconcile these myself. I suspect the best thing to do here is going to be to make separate commits for the different reconciliation approaches — i.e. do a 'twitter' to 'twitter' reconciliation first, and then commit the matches from that; then move on to reconciling by name, and commit the easy cases from that pass; then any more complex cases that you have to investigate or do more than a cursory check on, commit separately.

Then it will be much simpler to investigate where any differences are coming from, and decide what to do with them.

PR #14894, opened 2016-08-01
Turkey (Assembly): Add Twitter list scraper to sources 

octopusinvitro:
Scraper here: https://morph.io/everypolitician-scrapers/twitter-list-turkish-politicians

In this PR, I'm trying a new approach described [here](https://github.com/everypolitician/everypolitician-data/pull/14489#issuecomment-236424056)
- **First pass that checked in the ones which matched by twitter handle**. In this case, I can't see a Twitter handle on the right-hand side of `twitter.html`, so I'll first commit the reconciliation results for cases with 100% confidence.
- **Then a second pass for people where I was very confident they're the same person**. Here I'll commit the results for cases with same names.
- Then, for ones that actually required research to decide (e.g. between multiple members with the same name), I'd be inclined to even **make individual commits with explanations in the commit messages**. 

I'm deferring the harder cases to #14903


octopusinvitro:
For the fixup I followed these to decide to add a member:
* Same name but it has no spaces, 
* Same name but name or surname or both are in CAPSLOCK, 
* Same name but a letter is missing a modifier (for example **Özdogan**/**Özdoğan** for **Gökcen Özdogan Enç**),
* Same name but it has a title, for example **Dr. Dr.Abdullah ÇALIŞKAN**/**Abdullah Çalışkan**,
* Same name but it includes the Parti name, for example **SerminBalık AK Parti**/**Sermin Balık**

For the cases where there were many people with the same name, I decided to let them in when both sides of `twitter.html` showed both the same name and the same person in the picture, and jumped those who didn't have a picture. For example:
* **Hüseyin Çelik** [hc_huseyincelik](https://twitter.com/hc_huseyincelik) has two matches, but there is a picture in one of them.


I also let in:
* **Yıldırım Mehmet Ramazanoğlu** based on a Google search, Wikipedia and what I think is a Turkish government site, and
* **Ayla Akat Ata** based on same searches and her Twitter background picture.

tmtmtmtm:
I'm pretty sure that https://twitter.com/Avmehmetcan (created 2011) doesn't belong to https://tr.wikipedia.org/wiki/Mehmet_Can (died 2010)…

octopusinvitro:
@tmtmtmtm :cry: :sob: :crying_cat_face:

octopusinvitro:
Ok, since this data seems to be harder that expected, I leave it for now and will come back to it later, as I sate in issue #14903

octopusinvitro:
Closing as the reconciliation didn't work

PR #15105, opened 2016-08-05
South Korea (National Assembly) new term20

### LINKED ISSUE #14151: South Korea: 20th Assembly
# Source of New Data

## Legislature

Assembly  (20th term)

## Source

[Team POPONG](http://popong.com/)'s spreadie: https://github.com/teampopong/data-assembly/blob/master/assembly.csv
yay!

## Type of Data

List of Members

## Notes

Via email from Hoony

davewhiteland:
Closes #14151, because it reintroduces [Team POPONG](http://popong.com/)'s spreadie https://github.com/teampopong/data-assembly/blob/master/assembly.csv as a source, which is now providing data for the 20th term.

That issue discussed how the data from that source was related to the 20th term, and the conclusion is that it _solely_ contains members of the 20th term (the data also indicates if the politician was also a member of previous terms, which is data we're not processing — since we were consuming 19th term data ([now frozen as a manual source CSV](https://github.com/everypolitician/everypolitician-data/commit/659aaae0becd88aede4ad22222d37b9d344b4e54) from that source, we _should_ have previous members in the 19th term data already).

This is a data PR adding members for the 20th term to South Korea — primarily using POPONG as the source. Specifically, the scraper is the same as we were using for term 19 (since the POPONG spreadsheet and API are now providing term 20 data). So this PR is really reinstating the scraper that was being used as a source for term 19 (now it's explicitly returning a `term` of 20 in its data).

Expected result: a new term, with new members. Existing members reconciled.


davewhiteland:
steady on @everypoliticianbot, I'm reconciling with the previous term first

davewhiteland:
So this appears to be adding 155 members to a house of 300 -- this _seems_ a lot but actually might not be completely off; I haven't managed to confirm accurate numbers from other sources, partly because news reporting focusses on the seats that change, and that's not necessarily a count of new people (e.g., a new person can occur in a seat that hasn't changed). But South Korea's 2016 legislative elections _were_ indeed disruptive to the status quo, and the data source (POPONG) is sound, so I propose merging this and refining it, especially once we have a better understanding of the relationship between the different data available from POPONG.

tmtmtmtm:
@davewhiteland This looks good in terms of the general set-up and I get the same results as you when reconciling. But I'm a little worried by the 5 or 6 people who share a POPONG identifier as someone who is obviously a different person, e.g.

screen shot 2016-08-11 at 13 12 00

or

screen shot 2016-08-11 at 13 12 56

This means that we're ending up pulling in bad data for those people. For example, we end up with two people in the output with a facebook handle of "jhkimdragon". https://www.facebook.com/jhkimdragon is fairly clearly the person on the right hand side above — i.e. is correct in the _existing_ data, and incorrect in the new data. 

My suspicion here is that this is because the scraper is getting the wrong results when [looking people up](https://github.com/everypolitician-scrapers/korea-popong-data/blob/545b923d87302bbb21f3ba9b69b7bdeeb5df8709/scraper.rb#L18-L37) using the POPONG API

chrismytton:
Closing in favor of https://github.com/everypolitician/everypolitician-data/pull/18445

PR #15323, opened 2016-08-08
Split out a 'base' Rubocop configuration

tmtmtmtm:
Rubocop can inherit configuration from a file at a remote URL. This
allows us to use the config in this repo, as the base for _all_ our
related ones, rather than having to keep them all in sync manually.

So split out the inheritable one to `.rubocop_base.yml`, and keep the
project specific config on the main .`rubocop.yml`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,47 +3,6 @@ AllCops:
     - 'data/**/*'
     - 'vendor/**/*'
 
-Lint/AssignmentInCondition:
-  Enabled: false
-
-Lint/Debugger:
-  Enabled: false
-
-Style/AlignHash:
-  EnforcedHashRocketStyle: table
-  EnforcedColonStyle: table
-
-Style/ClassAndModuleCamelCase:
-  Enabled: false
-
-Style/CollectionMethods:
-  Enabled: true
-
-Style/Documentation:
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -3,47 +3,6 @@ AllCops:
     - 'data/**/*'
     - 'vendor/**/*'
```
chrismytton: I wonder if this line should be in `.rubocop_base.yml` as well? That would avoid problems with Travis installing gems into `vendor/`.

chrismytton:
Just one comment, other than that it looks good to me! 👍

.rubocop.yml
tmtmtmtm: It's already in the config that comes with Rubocop by default. 

We discovered last week, however, that if you supply your own Exclude list at all, it completely clobbers the built-in one, rather than extending it. So you need to bring back the defaults manually that you want from that. 

In other words, if you don't have any Excludes, it'll automatically skip 'vendor', so I don't think we need it in _our_ defaults. We only need to re-include any time we have any project-specific Excludes.

chrismytton: Ah OK, that makes sense. In that case you can ignore my previous comment :)

PR #15414, opened 2016-08-10
Fix reconciler match highlighting

chrismytton:
This was broken in a previous pull request because I forgot the `return` from the `_.any` function body.

Fixes https://github.com/everypolitician/everypolitician/issues/478

screen shot 2016-08-10 at 11 31 04


davewhiteland:
Ah right, JavaScript not explicitly returning anything is always making the comparison false.

So that looks good to me

tmtmtmtm:
👍

PR #15419, opened 2016-08-10
Singapore (Parliament): refresh data

everypoliticianbot:
Automated data refresh for Singapore - Parliament

#### Output

```
⤈ No ORDER BY for sources/morph/data.csv
⤈ No ORDER BY for sources/morph/terms.csv
Add memberships from sources/morph/data.csv
Merging with sources/morph/wikidata.csv
* 3 of 138 unmatched
	{:id=>"Q16136702", :name=>"Lina Loh"}
	{:id=>"Q314978", :name=>"Ali ibn al-Husayn Zayn al-'Abidin"}
	{:id=>"Q24006711", :name=>"Murali Pillai"}
Adding GenderBalance results from sources/gender-balance/results.csv
  ⚥ data for 0; 0 added

No dates for election Q20983901 (Next Singaporean general election)
Creating 13 term files

Top identifiers:
  135 x wikidata
  16 x viaf
  10 x freebase
  6 x sudoc
  5 x lcauth

Creating names.csv
Creating unstable/positions.csv
  ☇ No dates for S Jayakumar (Q1521378) as Emeritus Senior Minister
  ☇ No dates for S Jayakumar (Q1521378) as Co-ordinating Minister for National Security
  ☇ No dates for S Jayakumar (Q1521378) as Deputy Prime Minister of Singapore
  ☇ No dates for S Jayakumar (Q1521378) as Minister for Foreign Affairs
  ☇ No dates for Wong Kan Seng (Q461128) as Deputy Prime Minister of Singapore
  ☇ No dates for Wong Kan Seng (Q461128) as Co-ordinating Minister for National Security
  ☇ No dates for Wong Kan Seng (Q461128) as Minister for Home Affairs
  ☇ No dates for Wong Kan Seng (Q461128) as Minister for Foreign Affairs
  Unknown position (x1): Q3024268 Ministry of Communications and Information — e.g. Q5070296
  Unknown position (x1): Q2382270 Ministry of Community Development, Youth and Sports — e.g. Q5070296
  Unknown position (x1): Q4294530 Ministry of Education — e.g. Q983752
  Unknown position (x1): Q4383964 Ministry of Health — e.g. Q58142
  Unknown position (x1): Q15151135 Ministry of Social and Family Development — e.g. Q5070296
  Unknown position (x1): Q4294727 Ministry of Trade and Industry — e.g. Q58142
  Unknown position (x1): Q2289041 Singapore Democratic Party — e.g. Q1071599
[...]

PR #15510, opened 2016-08-12
Revert "Singapore (Parliament): refresh data"

octopusinvitro:
Sometimes octopuses brains malfunction as if they had fallen into a barrel full of booze.

Reverts everypolitician/everypolitician-data#15419


tmtmtmtm:
> shouldn't be merged because it was too old! A reminder to check the dates better.

@octopusinvitro — the problem with the PR had nothing to do with its age. It shouldn't have been merged because it lost lots of data, and introduced lots of duplicate people.

PR #15487, opened 2016-08-12
Singapore (Parliament): refresh data

everypoliticianbot:
Automated data refresh for Singapore - Parliament

#### Output

```
⤈ No ORDER BY for sources/morph/data.csv
⤈ No ORDER BY for sources/morph/terms.csv
Add memberships from sources/morph/data.csv
Merging with sources/morph/wikidata.csv
* 3 of 138 unmatched
	{:id=>"Q314978", :name=>"Ali ibn al-Husayn Zayn al-'Abidin"}
	{:id=>"Q24006711", :name=>"Murali Pillai"}
	{:id=>"Q16136702", :name=>"Lina Loh"}
Adding GenderBalance results from sources/gender-balance/results.csv
  ⚥ data for 0; 0 added

No dates for election Q20983901 (Next Singaporean general election)
Creating 13 term files

Top identifiers:
  135 x wikidata
  16 x viaf
  10 x freebase
  6 x sudoc
  5 x lcauth

Creating names.csv
Creating unstable/positions.csv
  ☇ No dates for S Jayakumar (Q1521378) as Emeritus Senior Minister
  ☇ No dates for S Jayakumar (Q1521378) as Co-ordinating Minister for National Security
  ☇ No dates for S Jayakumar (Q1521378) as Deputy Prime Minister of Singapore
  ☇ No dates for S Jayakumar (Q1521378) as Minister for Foreign Affairs
  ☇ No dates for Wong Kan Seng (Q461128) as Deputy Prime Minister of Singapore
  ☇ No dates for Wong Kan Seng (Q461128) as Co-ordinating Minister for National Security
  ☇ No dates for Wong Kan Seng (Q461128) as Minister for Home Affairs
  ☇ No dates for Wong Kan Seng (Q461128) as Minister for Foreign Affairs
  Unknown position (x1): Q3024268 Ministry of Communications and Information — e.g. Q5070296
  Unknown position (x1): Q2382270 Ministry of Community Development, Youth and Sports — e.g. Q5070296
  Unknown position (x1): Q4294530 Ministry of Education — e.g. Q983752
  Unknown position (x1): Q4383964 Ministry of Health — e.g. Q58142
  Unknown position (x1): Q15151135 Ministry of Social and Family Development — e.g. Q5070296
  Unknown position (x1): Q4294727 Ministry of Trade and Industry — e.g. Q58142
  Unknown position (x1): Q2289041 Singapore Democratic Party — e.g. Q1071599
[...]

PR #15514, opened 2016-08-12
Singapore add missing dates

octopusinvitro:
This PR puts back the missing `start_date` fields for those who lost them.

There were also some people who was renamed on the page and was given a new id based on their new names, so I remapped the old UUIDs to the new ids from the page.

This PR fixes #15487 which can be deleted, which was build on a bad merging made by some crazy octopus from which we [already recovered](https://github.com/everypolitician/everypolitician-data/pull/15510) thanks to the magic of git.


tmtmtmtm:
@octopusinvitro This claims to put back the missing start dates… but the diff shows that they're all still being removed:

screen shot 2016-08-13 at 05 57 59

You're also getting a lot of diffs relating to the `source` field for people. This isn't particularly _important_ (in fact I'm going to open a new PR to remove them for this legislature, as they're pointing at list pages, rather than person pages), but it's a warning sign that the data file you're working from is probably different from the one on Morph. I think you'd be much better just using the data from there, rather uploading an export from your own sqlite.

tmtmtmtm:
superseded by #15627 (but including the remapping from here)

PR #16057, opened 2016-08-25
Fill in missing area names when combining sources

chrismytton:
This adds some logic to combine sources which checks for any rows that have an OCD-like `:area_id` but no `:area` column, then tries to populate the `:area` column using the name from the OCD CSV.


tmtmtmtm:
The general approach here looks OK to me. However, I'm scared by this adding even more complexity to `combine_sources` when we're trying to simplify that. So finding a way to add this same functionality whilst also _reducing_ the `flog` score for the method would be nice.

(Also worth including a rebuild for Uganda here, as that seems to be the only place this actively changes)

PR #16061, opened 2016-08-25
Add `flog` to Gemfile

tmtmtmtm:
We want to reduce the complexity of the build process, so tracking the [Flog](http://ruby.sadi.st/Flog.html) scores is a good proxy for checking if we're moving in the right direction.


chrismytton:
👍

PR #16064, opened 2016-08-25
Require ruby 2.1.0 onwards

tmtmtmtm:
support for ruby 2.0.0 (including security fixes) ended in February:
 https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

Upgrading our minimal version to 2.1 also gives us aaccess to a few features we
want to start using, including:
* Pathname.write
* required named parameters


chrismytton:
Woo! 👍

PR #16078, opened 2016-08-25
Use full commit hash rather than abbreviated one

tmtmtmtm:
We recently hit a case where the short version doesn't work: Peru was pointing at https://github.com/everypolitician/everypolitician-data/commit/8b15cc3, but we have a clash between https://github.com/everypolitician/everypolitician-data/commit/8b15cc3c (which is what we need here) and https://github.com/everypolitician/everypolitician-data/commit/8b15cc39 

So instead switch to using the full version.


chrismytton:
LGTM! 🚀

PR #16239, opened 2016-08-30
Switch Australia to use Wikidata area ids

chrismytton:
This uses area information scraped from Wikidata in https://morph.io/everypolitician-scrapers/australia-constituencies-wikidata to generate an `:area_id` column which contains a wikidata area rather than an OCD id.


tmtmtmtm:
https://github.com/everypolitician/everypolitician-data/pull/51981 added Wikidata information for each area. We'd still like to remove the OCD IDs, but we'd do that now by dropping that source re-reconciling the Wikidata IDs against the underlying IDs instead.

PR #16368, opened 2016-09-02
Extract a StatsFile class

tmtmtmtm:
As part of the process of moving logic out of the Rakefile, and into
classes, pull the body of the `generate_stats` task out into an
(inlined) StatsFile class. There’s still much more to do on this,
including simplifying some of the logic, and moving it out to its own
file (and quite possibly renaming it, depending on whether it continues
to only represent the data hash, rather than the file itself), but this
should be a good start.


tmtmtmtm:
This currently fails to run cleanly for Åland, dying with 
```
NameError: undefined local variable or method `identifiers' for #<Everypolitician::Popolo::Organization:0x007f943c7a9810>
```

tmtmtmtm:
@chrismytton this is the same problem we've had before, where if an Organization doesn't actually _have_ any identifiers, the method simply doesn't exist, making it essentially impossible to write code even as simple as `popolo.organisations.each { |o| puts "#{p.id}: #{p.wikidata}" }` without lots of defensive `rescue` or `respond_to?` type stuff. I say we take that whole `define_singleton_method` approach out back and dispose of it with extreme prejudice. We know what the methods are that should exist on each of these, so let's just create them properly.

tmtmtmtm:
@ondenman I've extracted this all out into its own class. Can you work with @chrismytton on making this work (in the first case it's probably a matter of just adding a proper `identifiers` method in `everypolitician-popolo`, though we should find a better approach generally), add some tests against a known Popolo file (I'd suggest for Åland, as it's one that's definitely failing atm), and do any final tidy-ups you think are required.

tmtmtmtm:
@ondenman @chrismytton I needed the fix to `everypolitician-popolo` elsewhere as well, so I made a new release of that, and included it here. There are other things that could be done to tidy this one up, but to reduce the risk of conflicts, I'm going to merge it now. You should still continue the related work on both directions of this (improvements to `everypolitician-popolo` so we can use it for all these generations, whilst also extracting the logic of them from the Rakefile to library files) next week.

Issue #16435, opened 2016-09-02
Broken Import: Guyana


tmtmtmtm:
## Legislature

Guyana - National Assembly

## Problem

Losing lots of prefix and gender information — https://github.com/everypolitician/everypolitician-data/pull/16362/files

PR #16581, opened 2016-09-06
16435 guyana reinstate missing titles

ondenman:
Mr/Mrs titles have been removed from some member pages. Since the scraper derives gender info from these titles, members were losing gender info.

Example:
screen shot 2016-09-06 at 15 21 01

screen shot 2016-09-06 at 15 20 22

Titles and gender have been added to manual files.

Closes https://github.com/everypolitician/everypolitician-data/issues/16435


tmtmtmtm:
I don't think we want the entries in this that are blank, e.g.:

```diff
+bharrat-jagdeo,`
```

I'm also a little surprised you did this as two new sources, rather than as a corrections file, but i think that works OK, as it assigns source info. I'd make the commit message for each commit that adds the data a little more verbose though, for the benefit of someone trying to track down where this data came from months or even years from now.

tmtmtmtm:
@octopusinvitro it would still be good to remove the needless blank entries from these two files.

PR #16592, opened 2016-09-06
Falkland Islands (Legislative Assembly): refresh data

everypoliticianbot:
Automated data refresh for Falkland Islands - Legislative Assembly

#### Output

```
Add memberships from sources/morph/wikipedia.csv
Merging with sources/morph/official.csv
Merging with sources/morph/wikidata.csv
Adding GenderBalance results from sources/gender-balance/results.csv
  ⚥ data for 75; 36 added

No dates for election Q15744904 (Next Falkland Islands general election)
Creating 10 term files

Top identifiers:
  20 x wikidata
  1 x clara
  1 x viaf

Creating names.csv
Creating unstable/positions.csv
Persons matched to Wikidata: 20 ✓ | 30 ✘
  No wikidata: Harold Rowlands (f0c31a11-1ad4-4d7a-8bf6-eca28b529d4a)
  No wikidata: John Edward Cheek (c636cd6f-fb28-4606-8891-a2ebd02b08dc)
  No wikidata: Terence John Peck (5d427660-db2d-4b3f-86f0-314a99cb64ad)
  No wikidata: Stephen Luxton (d8de53b7-0798-455a-bba0-4e77b61d2017)
  No wikidata: Kevin Kilmartin (3269f76c-cf87-4e6e-859d-122f91fb9132)
  No wikidata: Stuart Barrett Wallace (8de4e6a8-bd09-43a8-a979-520678f2366c)
  No wikidata: Lewis Clifton (81ab01d9-2345-4262-84aa-00014c29bda4)
  No wikidata: Lionel Geoffrey Blake (63fae433-6182-4ff2-a9cb-3846b8b8a57f)
  No wikidata: Ron Binnie (9cf9611e-8877-4f1f-a613-1912bac64424)
  No wikidata: William Henry Goss (72a1c546-2e4d-4919-9669-6bcb59fcd5e2)
Parties matched to Wikidata: 1 ✓ 

```

PR #16718, opened 2016-09-08
Run pull request cleaner on Travis

chrismytton:
This adds a new Rake task that gets run on Travis which goes through the open pull requests and closes any that are outdated. It uses [close_old_pull_requests](https://github.com/everypolitician/close_old_pull_requests) to do the actual work, and it replaces [pull_request_cleaner](https://github.com/everypolitician/pull_request_cleaner).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: ruby
+script: bundle exec rake test
 rvm:
   - 2.3.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,6 @@ language: ruby
 script: bundle exec rake test
 rvm:
   - 2.3.1
+env:
+  global:
+    secure: ZCQVQdKDUyPY/BM1FYriv6h+rNF+PqQETnj5FwU/jQKNz8utvaUVjXD/pIMwEmO5orzng5/MnZvErkr2+JZBs+0JghXWtjDxQHxJzfBTZJnaiAY9hWZBC9Kgbl1/mJvUvK132W5GzQ5xS9lw/pyKFGRba9A7rfoYFfkaJKmgq1U=
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-script: bundle exec rake test
+script: bundle exec rake test close_old_pull_requests
 rvm:
   - 2.3.1
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 script: bundle exec rake test close_old_pull_requests
+cache: bundler
 rvm:
   - 2.3.1
 env:
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem 'json5'
 gem 'slop', '~> 3.6.0' # tied to pry version
 gem 'rcsv'
 gem 'require_all'
+gem 'close_old_pull_requests', git: 'https://github.com/everypolitician/close_old_pull_requests', branch: 'master'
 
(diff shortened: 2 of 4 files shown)


tmtmtmtm:
@chrismytton I find the output from this a little confusing. When I ran it it reported:

```
Pull request 16560 is outdated. (Newest pull request is 16748)
Pull request 16592 is outdated. (Newest pull request is 16738)
Pull request 16678 is outdated. (Newest pull request is 16721)
```

This makes it look like I then need to do something else to deal with these. It would be better if this could report on what happened (i.e. that it closed them, or wasn't able to to, etc)

I'm OK with this being tidied up after it goes live, due to the time pressure on this, but a little more seriously, be it also doesn't actually seem to work (at least in some cases)

https://github.com/everypolitician/everypolitician-data/pull/16678 got a 

```
This Pull Request has been superseded by #16721 but there are non-bot commits.

@everypoliticianbot is this pull request still needed?
```
but there doesn't actually appear to _be_ any non-bot commits there.

I'm also curious about what's going to happen in cases where there actually _are_ non-bot commits? e.g. is https://github.com/everypolitician/everypolitician-data/pull/16592 going to get hundreds of comments every day asking if it should still exist?

.travis.yml
```diff
@@ -1,3 +1,8 @@
 language: ruby
+script: bundle exec rake test close_old_pull_requests
```
tmtmtmtm: this makes it look like we're adding `rake test` as well as `rake close_old_pull_requests`. I'm assuming that there was just an implied `script: bundle exec rake test` previously because we didn't supply any line at all? If so it might be useful to make that version explicit first as a separate commit, so that it's more obvious that _this_ commit is only adding the `close_old_pull_requests` part.

chrismytton:
> @chrismytton I find the output from this a little confusing. When I ran it it reported:
> 
> Pull request 16560 is outdated. (Newest pull request is 16748)
> Pull request 16592 is outdated. (Newest pull request is 16738)
> Pull request 16678 is outdated. (Newest pull request is 16721)
> This makes it look like I then need to do something else to deal with these. It would be better if this could report on what happened (i.e. that it closed them, or wasn't able to to, etc)

Yep good point, ticketed as https://github.com/everypolitician/close_old_pull_requests/issues/7.



> I'm OK with this being tidied up after it goes live, due to the time pressure on this, but a little more seriously, be it also doesn't actually seem to work (at least in some cases)
> 
> #16678 got a
> 
> This Pull Request has been superseded by #16721 but there are non-bot commits.
> 
> @everypoliticianbot is this pull request still needed?
> but there doesn't actually appear to be any non-bot commits there.

Ah, that's a bug where I've assumed that the GitHub access token will always belong to the bot. Ticketed as https://github.com/everypolitician/close_old_pull_requests/issues/6.



> I'm also curious about what's going to happen in cases where there actually are non-bot commits? e.g. is #16592 going to get hundreds of comments every day asking if it should still exist?

Hmm, hundreds of comments a day is not ideal is it! Ticketed as https://github.com/everypolitician/close_old_pull_requests/issues/5.

chrismytton:
> Perhaps worth adding a pointer in the commit message to how the GITHUB_ACCESS_TOKEN environment variable actually got turned into the line that's actually in the file? (Presumably just by adding a URL that describes that process)

Ah yep, will do.

chrismytton:
Merging this now. I've fixed all the close_old_pull_request_issues except https://github.com/everypolitician/close_old_pull_requests/issues/7, which we can improve once it's live. Then I'll turn off the cleaner on Heroku so this version can run over the weekend.

PR #16773, opened 2016-09-09
Guyana (N. Assembly) Remove people with no known gender

octopusinvitro:
The gender.csv manual file was added to recover the gender
info removed from the upstream sources. But if we also
add people here for which we don't know the gender yet, we
are stating that they actually don't have one. So this commit
removes people with gender unknown.


chrismytton:
LGTM! 👍  🎸

PR #16774, opened 2016-09-09
 Guyana (N. Assembly) Remove people with no honorific title

octopusinvitro:
The honorific_prefix manual file was added to recover the titles
that were removed from the upstream sources. But if we also
add people here for which we don't know the title yet, we are
stating that they actually don't have one. So this commit
removes people with no title.


chrismytton:
LGTM! 👍  :metal:

PR #16985, opened 2016-09-12
Run pull request summarizer on Travis CI

chrismytton:
This adds an extra rake task to the Travis CI build. When the task is run it will add a comment to the current pull request containing a summary of changes.

Similar to #16718 
Related to https://github.com/everypolitician/everypolitician/issues/449

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -67,3 +67,24 @@ task :close_old_pull_requests do
     puts "Pull request #{pull_request.number} is outdated. (Newest pull request is #{pull_request.superseded_by.number})"
   end
 end
+
+desc "Post a summary of the pull request that's currently being built"
+task :pull_request_summary do
+  if ENV['TRAVIS_PULL_REQUEST'] == 'false'
+    warn 'Not building a pull request, skipping pull_request_summary'
+    next
+  end
+  require 'everypolitician/pull_request'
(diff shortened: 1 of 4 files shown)


Rakefile.rb
```diff
@@ -67,3 +67,13 @@ def json_write(file, json)
     puts "Pull request #{pull_request.number} is outdated. (Newest pull request is #{pull_request.superseded_by.number})"
   end
 end
+
+desc "[Travis only] Post a summary of the pull request that's currently being build"
```
tmtmtmtm: typo "build" → "built"

chrismytton: Gah, every time! This typo will be with me for life 😢

Rakefile.rb
```diff
@@ -67,3 +67,24 @@ def json_write(file, json)
     puts "Pull request #{pull_request.number} is outdated. (Newest pull request is #{pull_request.superseded_by.number})"
   end
 end
+
+desc "Post a summary of the pull request that's currently being built"
+task :pull_request_summary do
+  if ENV['TRAVIS_PULL_REQUEST'] == 'false'
```
tmtmtmtm: I think this would all be a lot cleaner if we had two separate tasks: one for travis to run, and one for humans to run.

chrismytton: Yep, good point. I'll merge this now so I can shut off the Heroku bits, then split that up as a separate pull request.

tmtmtmtm:
dancing_penguins

PR #16987, opened 2016-09-12
Use rake task from Everypolitician::PullRequest

chrismytton:
~~This change splits up the `rake pull_request_summary` command. There are now three versions:~~

~~1. `pull_request_summary:print` will print a summary of the pull request
    to the terminal.
2. `pull_request_summary:comment` is for running locally and will post a
    comment to GitHub as you if there's an access token in the environment.
3. `pull_request_summary:travis` is for running on Travis and will post
    a comment to GitHub as the bot.~~

This switches to the rake task that was introduced in https://github.com/everypolitician/everypolitician-pull_request/pull/6. This means that we're not filling up the rakefiles with code that's better suited living in the same place as the library it calls.

Follows on from #16985

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-script: bundle exec rake test close_old_pull_requests pull_request_summary
+script: bundle exec rake test close_old_pull_requests pull_request_summary:travis
 cache: bundler
 rvm:
   - 2.3.1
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
Yay. This is a big improvement. I'm a little concerned though that we're polluting the Rakefiles with too much logic, and that having slimmed this one down substantially in the last month or so, we've effectively doubled the size of it again in the last few days. 

Perhaps simply putting these tasks in a Rakefile of their own might be a suitable next step?

tmtmtmtm:
Wooot! 
penguin-dance

PR #16921, opened 2016-09-11
Ukraine (Verkhovna Rada): refresh data

everypoliticianbot:
Automated data refresh for Ukraine - Verkhovna Rada

#### Output

```
/tmp/d20160911-3-47k6jf/vendor/bundle/bundler/gems/everypolitician-popolo-d1175550b752/lib/everypolitician/popolo.rb:69: warning: already initialized constant EveryPolitician
/tmp/d20160911-3-47k6jf/lib/country_data.rb:2: warning: previous definition of EveryPolitician was here
Wikidata renamings:
	Q20067945
Add memberships from sources/morph/data.csv
Merging with sources/morph/wikidata.csv
* 6 of 456 unmatched
	{:id=>"Q1708369", :name=>"Одарченко Юрій Віталійович"}
	{:id=>"Q24958301", :name=>"Куліч Валерій Петрович"}
	{:id=>"Q22953373", :name=>"Микитась, Максим Вікторович"}
	{:id=>"Q25438057", :name=>"Ричкова Тетяна Борисівна"}
	{:id=>"Q7380984", :name=>"Богдан Руслан Дмитрович"}
	{:id=>"Q17372415", :name=>"Шахов Сергій Володимирович"}
Adding GenderBalance results from sources/gender-balance/results.csv
  ⚥ data for 292; 0 added

Creating 1 term file

Top identifiers:
  458 x rada
  450 x wikidata
  37 x freebase
  32 x viaf
  15 x lcauth

Creating names.csv
Creating unstable/positions.csv
  Unknown position (x1): Q30185 mayor — e.g. Q4150130
  Unknown position (x1): Q212071 rector — e.g. Q20070496
  Unknown position (x1): Q12164846 Commissioner for anti-corruption policy — e.g. Q12170540
Persons matched to Wikidata: 450 ✓ | 8 ✘
  No wikidata: Одарченко Юрій Віталійович (0b1dca15-008b-49f7-bb61-185479544ad3)
  No wikidata: Шевченко Віктор Леонідович (f75e4346-07f4-435d-b12b-fba7dcc44cc0)
  No wikidata: Констанкевич Ірина Мирославівна (16715574-db3c-49bf-bb16-e3aacd13e10c)
  No wikidata: Шахов Сергій Володимирович (44c45b5a-2af0-4f6c-bf34-0ef3f0f6840c)
  No wikidata: Микитась Максим Вікторович (b3bb9791-523b-4e17-b2c5-b3b967dd85ee)
[...]

PR #16988, opened 2016-09-12
Add rake task to handle moved Wikidata

tmtmtmtm:
if a Wikidata page is moved, e.g. https://www.wikidata.org/w/index.php?title=Q20067945&redirect=no being renamed to https://www.wikidata.org/wiki/Q24958301, we need to update our reconciliation file to point the new ID at our UUID.

This adds a rake task to do that. You call it from the legislature directory as:

``` 
rake wikidata:handle_move[Q20067945,Q24958301]
```

(some shells, e.g. zsh, may need to quote those square brackets)

## Notes to Reviewer

1. Making this rake task simple required adding a class to represent the reconciliation file. We should also switch lots of the existing reconciliation code to _use_ this new class, but that should happen later.

2. I’m not entirely happy with the name of the rake task (`handle_move`). I'm very open to alternative suggestions for it!


tmtmtmtm:
This was used to make https://github.com/everypolitician/everypolitician-data/pull/16921/commits/8831d1ef99925c548ac1f2b75cb743b035a83b86

chrismytton:
> I’m not entirely happy with the name of the rake task (handle_move). I'm very open to alternative suggestions for it!

What about `rake wikidata:handle_redirect` or `rake wikidata:handle_rename`?

Other than potentially renaming the rake task this looks good to me! 👍  🚀

tmtmtmtm:
It's more the `handle_` part that I'm less convinced by… I think let's merge for now, and see if a better name pops up later.

@davewhiteland you may wish to update some wiki documentation for this.

cc: @octopusinvitro / @ondenman FYI

PR #16560, opened 2016-09-06
Dominican Republic (Chamber of Deputies): Add new term

everypoliticianbot:
This PR adds the new 2016 term and moves the 2010 term to the manual sources since it seems to have disappeared from the official sources.

The bot commits have been deleted .

PR #17006, opened 2016-09-13
Dominican Republic: Move the 2010 term to manual files

octopusinvitro:
Both the official site and wikipedia seem to be showing only the
data for the 2016 term, so at the moment morph scraper is mixing
the data already available from term 2010 with the new incoming
data for term 2016. This commit moves morph/data.csv to
maunal/term-2010.csv so that the morph database can be deleted
and regenerated with just the data from 2016 term.

When running `bundle exec rake clean default` no files were changed.

This PR is the first part of splitting #16560


tmtmtmtm:
Ah, no, I spoke too soon (and can't see a way in this new interface to undo a review). This doesn't work, as it leaves behind the old stanza in the `instructions.json` as well, so when running a `bundle exec rake clean default` _lots_ of files actually change.

PR #17335, opened 2016-09-18
use everpolitician-dataview-terms to generate CSVs

tmtmtmtm:
The code for generating the term CSVs has been extracted to its own gem:
  https://github.com/everypolitician/everypolitician-dataview-terms

This replaces the internal implementation with calls to that library instead.

`bundle exec bin/run_command_for_each_legislature rake clean default` shows no changes to any CSV file.

Part of https://github.com/everypolitician/everypolitician/issues/320


tmtmtmtm:
@chrismytton for sanity checking. If this is a sensible approach, then we can start to pull out the code for generating the other post-Popolo files.

PR #17400, opened 2016-09-19
EP_COUNTRY_REFRESH should work by slug

tmtmtmtm:
## Context

Previously we assumed we were working with a country in directory-name format (e.g. South_Africa), but the Rebuilder is passing us slugs (i.e. South-Africa). That's the more correct behaviour, so switch to assuming that's what we're getting.

## Related tickets

Closes https://github.com/everypolitician/everypolitician/issues/486

## Notes to reviewer
The implementation is a little bit clumsy at the moment — we can clean things up substantially by making everything assume they'll be dealing with `EveryPolitician::Country` or `EveryPolitician::Legislature objects`, but that should happen as a separate change.

Also, there's still a slight bootstrapping problem here when we add a new country, but in that case we can just generate an entire file, rather than by country name.

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -1,4 +1,5 @@
 
+require 'everypolitician'
 require 'fileutils'
 require 'pathname'
 require 'pry'
@@ -26,9 +27,8 @@ task 'countries.json' do
   # we only build any country that contains that string. For example:
   #    EP_COUNTRY_REFRESH=Latvia be rake countries.json
   to_build = ENV['EP_COUNTRY_REFRESH'] || 'data'
-
-  countries = @HOUSES.group_by { |h| h.split('/')[1] }.select do |_, hs|
-    hs.any? { |h| h.include? to_build }
+  countries = EveryPolitician.countries.select do |c|
+    c.slug.downcase.include? to_build.downcase
   end
 
   data = json_load('countries.json') rescue {}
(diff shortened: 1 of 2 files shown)


tmtmtmtm:
@chrismytton for review

Rakefile.rb
```diff
@@ -26,23 +27,25 @@ def json_write(file, json)
   # we only build any country that contains that string. For example:
   #    EP_COUNTRY_REFRESH=Latvia be rake countries.json
   to_build = ENV['EP_COUNTRY_REFRESH'] || 'data'
-
-  countries = @HOUSES.group_by { |h| h.split('/')[1] }.select do |_, hs|
-    hs.any? { |h| h.include? to_build }
+  countries = EveryPolitician.countries.select do |c|
+    c.slug.downcase.include? to_build.downcase
```
chrismytton: Using `.include?` here means that setting `EP_COUNTRY_REFRESH=Dominica` will _also_ rebuild the Dominican Republic. That's not necessarily a bad thing, but it could be unexpected. Perhaps this should either do an exact match _or_ print out the countries it's refreshing?

tmtmtmtm: That's the current behaviour, so removing it would be a big change, especially as it's behaviour that I often take advantage of from the command line (e.g. for countries that I can't always remember how to spell!) 

The goal of this argument is really about speed, not isolation, and the only penalty here is in saving a little _less_ time over the full rebuild than an exact match would do. 

In the case where this is triggered by a bot, and we can rely on an exact match, it might make sense to have a way to say so. But the extra value we'd gain from that doesn't really seem worth the extra overhead of having two arguments or a different way to flag that up. Definitely open to suggestions of how we could have both behaviours without any extra complexity, though!

Printing out the countries it's changing is tempting, though again it would be a change of behaviour if it suddenly had output by default, so we'd probably want to hide that behind a verbose flag or something, and again I'm not sure it's important enough. Open to persuasion though if you want to open a ticket for it.

PR #17407, opened 2016-09-19
Deploy everypolitician.org from Travis

chrismytton:
This adds a script to deploy http://everypolitician.org that runs each time Travis successfully builds the `master` branch.

It does this by installing [viewer-sinatra](https://github.com/everypolitician/viewer-sinatra) and using it to rebuild [viewer-static](https://github.com/everypolitician/viewer-static).

Related to https://github.com/everypolitician/everypolitician/issues/506 as part of [changing how everypolitician.org is deployed](https://github.com/everypolitician/everypolitician/projects/2).

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -21,7 +21,7 @@ build_viewer_static() {
 
 deploy_viewer_static() {
   cd /tmp
-  git clone https://github.com/everypolitician/viewer-static.git
+  git clone --depth=1 https://github.com/everypolitician/viewer-static.git
   cd viewer-static
   git checkout gh-pages
   cp -R /tmp/localhost:4567/* .
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+[[ "$TRACE" ]] && set -x
+
+start_viewer_sinatra() {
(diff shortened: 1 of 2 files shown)


chrismytton:
@tmtmtmtm 👀

scripts/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+build_viewer_static() {
+  cd /tmp
+  wget -nv -m localhost:4567/status/all_countries.html || (echo "wget exited with non-zero exit code: $?" >&2 && exit 1)
+}
+
+deploy_viewer_static() {
+  cd /tmp
+  git clone https://github.com/everypolitician/viewer-static.git
```
tmtmtmtm: Would we be better with a shallow clone here?

chrismytton: Ah yes, that's a very good idea. Amazingly, according to git, a shallow clone with `--depth 1` _still_ results in a 13.68 MiB clone, but I suppose that's much better than 700+MB! Will add that now.

chrismytton:
@tmtmtmtm I've addressed your comments so I think this is ready for another review now.

tmtmtmtm:
Excellent. We presumably also want to point back to this one from there as well — I'm assuming we'll also want to simplify that one to use the same deploy mechanism too, rather checking TRAVIS_PULL_REQUEST etc? (assuming so, we should probably have an Issue for that, if there isn't one already)

PR #17409, opened 2016-09-19
Map existing UUIDs to new incoming IDs

ondenman:
In a previous PR (https://github.com/everypolitician/everypolitician-data/pull/17115), the bot reported that many new members were being added. These members were existing members but had received new upstream IDs. This PR assigns their original UUIDs to their new upstream IDs.

Also removes duplicate Søren Pape Poulsens and Rune Lunds.

screen shot 2016-09-20 at 09 48 21

screen shot 2016-09-20 at 09 49 30


data/Denmark/Folketing/ep-popolo-v1.0.json
```diff
@@ -4660,14 +4660,14 @@
       "contact_details": [
         {
           "type": "email",
-          "value": "min@kum.dk"
+          "value": "min@kum.dk|km@km.dk"
```
tmtmtmtm: This doesn't look like a valid email address. I'm not sure where this is coming from, but you'll want to separate these with a semicolon rather than than a pipe, and then the build process will Do The Right Thing.

ondenman:
Breaking this up. Starting with https://github.com/everypolitician/everypolitician-data/pull/17478

PR #17441, opened 2016-09-20
Add missing switch to tar command in deploy script

chrismytton:
Travis was failing to deploy because the version of tar(1) on Linux
doesn't automatically add the -z switch like the OS X version does.

https://travis-ci.org/everypolitician/everypolitician-data/builds/161313383#L312-L313


chrismytton:
@tmtmtmtm 👀

PR #17451, opened 2016-09-20
Use .netrc on Travis for GitHub access token

chrismytton:
Rather than embed the access token in the url we want to add it to `.netrc` so that it gets picked up by git.

This is the same as the change I made in https://github.com/everypolitician/viewer-sinatra/pull/15551, it should have been part of https://github.com/everypolitician/everypolitician-data/pull/17407 but I forgot to include it 😞 

Travis docs: https://docs.travis-ci.com/user/private-dependencies/#API-Token


chrismytton:
@tmtmtmtm 👀

PR #17487, opened 2016-09-21
Try using a newer Ruby version for deploy script

chrismytton:
Travis [defaults to Ruby 1.9.3 for "reliability"](https://docs.travis-ci.com/user/deployment/script/#Deployment-is-executed-by-Ruby-1.9.3). This change means that the deploy script will be run in an [rvm environment](https://rvm.io/set/do) that configures the Ruby on the current `PATH` to be the same as the Ruby that we use for the rest of the build.

Fixes https://github.com/everypolitician/everypolitician/issues/511

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,6 @@ env:
     secure: ZCQVQdKDUyPY/BM1FYriv6h+rNF+PqQETnj5FwU/jQKNz8utvaUVjXD/pIMwEmO5orzng5/MnZvErkr2+JZBs+0JghXWtjDxQHxJzfBTZJnaiAY9hWZBC9Kgbl1/mJvUvK132W5GzQ5xS9lw/pyKFGRba9A7rfoYFfkaJKmgq1U=
 deploy:
   provider: script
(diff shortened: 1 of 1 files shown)


chrismytton:
@tmtmtmtm 👀

tmtmtmtm:
@chrismytton I think I'm missing some context here. This feels like there's a missing Issue somewhere that it can close…

chrismytton:
@tmtmtmtm Good point. It turns out that in the process of writing the issue I discovered that this fix is actually probably not what we want (though it's hard to tell, because testing Travis on the `master` branch is a bit tricky!). Anyway, the issue is https://github.com/everypolitician/everypolitician/issues/511, but this pull request isn't ready for review as I need to fix things in a different way.

chrismytton:
@tmtmtmtm I've updated the description to reference an issue and pushed an update so that this should now hopefully work correctly, so I think it's ready for review again.

chrismytton:
Have now added a comment to `.travis.yml` with a brief explanation of how the deploy script line is working.

PR #17489, opened 2016-09-21
Add rake task for recovering vanishing source data

tmtmtmtm:
A semi-regular pattern is that a morph scraper database gets emptied, but there were people in it that the scraper no longer picks up (because the data is no longer on the live site).

In those cases we extract the vanishing data from an older copy of the file, and add that as a "manual" source.

This is quite a fiddly process, so let’s have a rake task to do it for us.

This is part 1 of that, which rescues the rows that have gone missing from the actual data source.

We still need a part 2, which rescues the corresponding `idmap` entries, but we need https://github.com/everypolitician/everypolitician/issues/510 first

(There’s also a possible part 3 to add the manual file to instructions.json, but that can come even later)

With this in place we rescue data that has gone missing from the Isle of
Man scraper.

## Notes to reviewer

Very open to suggestions for better ways of getting the history than shelling out to 
```ruby
`git show @~1:#{full_pathname}`
```


tmtmtmtm:
@chrismytton  👀

rake_config/vanishing_data.rb
```diff
@@ -0,0 +1,74 @@
...
+  def repo_root
+    Pathname.new('../../..').realpath
+  end
+
+  def full_pathname
+    source.pathname.realpath.relative_path_from(repo_root)
+  end
+
+  def old
+    @old ||= SourceCSV.new(`git show @~1:#{full_pathname}`)
```
chrismytton: I can't immediately think of a better way to execute this command without shelling out to git. However I do think it would be clearer if the command used `HEAD` instead of `@` since I wasn't aware `@` was an alias for `HEAD` until I read `git help revisions` just now!

```ruby
`git show HEAD~1:#{full_pathname}`
```

rake_config/vanishing_data.rb
```diff
@@ -0,0 +1,74 @@
...
+  def initialize(str)
+    @raw = str
+  end
+
+  def ids
+    grouped.keys
+  end
+
+  def rows(*kk)
+    kk.map { |k| grouped[k].first }
```
chrismytton: Doesn't taking the `.first` item from the group mean we might potentially loose data if a membership CSV has multiple rows for a single member with the same id?

tmtmtmtm: Good catch. I think there are likely to be all sorts of odd edge cases when we're dealing with multiple memberships — e.g. only some of them disappearing. But I think we should defer handling those until we actually see them. This one is an easy fix though, so I've added it.

rake_config/vanishing_data.rb
tmtmtmtm: and deprive other people from learning that too? 😜

chrismytton: I guess the other option would be to move the whole git command to a separate method called `old_csv_contents` or something. But that can be improved on in the future. After discussing in Slack I think this is OK for now.

tmtmtmtm:
ping @octopusinvitro @ondenman @davewhiteland for awareness/wiki updates etc.

octopusinvitro:
I really like this task, automation FTW :+1: 
I know this is already merged, but I wanted to make some small observations.

Puts on moustache:

* This task is run as `bundle exec rake missing_data:memberships[members] > sources/manual/vanished.csv`. Do you think it would be a good idea to do the writing to a file through some class method when inside of the task instead of when we run the task? i.e., do `bundle exec rake missing_data:memberships[members]` and have the file written to disk?

* Small comments regarding naming:
  - I would have avoided cognitive overload in the constructor of `SourceCSV` by calling the argument `raw` instead of `str`. Or if sticking with a different name, using the whole word `string`, would be more readable. Or even better, a name that is more informative, `contents` maybe?

  - Same for `def rows(*kk)`. When I see this I have no idea what `kk` is (apart from a scatological word in Spanish :stuck_out_tongue_winking_eye: ) Would another person know what do they have to pass here?

  - Regarding `def cur`... what is the harm of writing `def current`? `cur` can be current, or can be a cursor (which is not crazy given that this class deals with files) or can be a curse. The curse that makes naming hard, for sure.

tmtmtmtm:
> Do you think it would be a good idea to do the writing to a file through some class method when inside of the task instead of when we run the task?

Probably not, for two main reasons: one practical and one philosophical.

I don't think the task can know where you're going to want to place the output. There is no connection between the name of file the data used to live in, and the name of the file where you're going to place it now. That's always something that's going to need to be supplied at run-time. 

Sure, you could always supply an extra argument to tell it that, but then you'd also need to write lots of code to cope with different conditions for whether that file already exists or not, etc. It's much simpler to adhere to the unix philosophy here and just spit out the output requested, and leave it up to the user to decide what to do with that: whether that's making a new file from it, adding it to the end of an existing one, or doing something else entirely with it which you shouldn't need to ever know about or care about.

> * Small comments regarding naming:

Yep. Good points. PRs welcome.

PR #16622, opened 2016-09-07
Germany (Bundestag): refresh data

tmtmtmtm:
```
⤈ No ORDER BY for sources/morph/data.csv
⤈ No ORDER BY for sources/morph/wikidata.csv
Add memberships from sources/morph/data.csv
Merging with sources/morph/wikidata.csv
* 8 of 3803 unmatched
	{:id=>"Q1647066", :name=>"Lothar Löffler"}
	{:id=>"Q1453729", :name=>"Heinz Meyer"}
	{:id=>"Q25906244", :name=>"Jürgen Coße"}
	{:id=>"Q916474", :name=>"Brigitte Lange"}
	{:id=>"Q1553109", :name=>"Gudrun Kopp"}
	{:id=>"Q15794238", :name=>"Christian Petry"}
	{:id=>"Q2562873", :name=>"Werner Weiß"}
	{:id=>"Q1710101", :name=>"Martin Neumann"}

Data Mismatches
  ☁ Mismatch in death_date for 1a73cd79-c798-42f5-bd15-75b514a9fa68 (2005-11-10) vs 2005-11-09 (for Q15427635)
  ☁ Mismatch in death_date for 7c874933-ebcc-4320-aed6-f199cc2fbe12 (1955-07-30) vs 1955-06-30 (for Q1438708)
  ☁ Mismatch in death_date for bb0fd1f9-8b6d-45cc-8e0e-8f5d40bbddc5 (1951-07-29) vs 1951-07-21 (for Q106049)
  ☁ Mismatch in death_date for 4f2c210b-b0a4-4db4-b9e3-a6be36e8bef3 (1952-04-22) vs 1952-04-23 (for Q2574416)
  ☁ Mismatch in death_date for 47c8e9a6-fd0d-40a8-926e-86849482b4fc (1950-11-21) vs 1950-10-21 (for Q1352568)
  ☁ Mismatch in death_date for f177b48a-6544-4a09-b68c-9395d25295ca (1996-02-06) vs 1996-02-07 (for Q1725753)
  ☁ Mismatch in death_date for edef6b9f-ab67-417a-9aa1-6c2e0b5ae280 (2001-05-10) vs 2001-05-09 (for Q98342)

Adding GenderBalance results from sources/gender-balance/results.csv
  ⚥ data for 1272; 25 added

Creating 18 term files

Top identifiers:
  3795 x wikidata
  2381 x viaf
  2203 x gnd
  1632 x bundestag
  862 x munzinger

Creating names.csv
[...]

PR #17571, opened 2016-09-23
Germany (Bundestag): refresh data.csv

tmtmtmtm:
Rebuild only a single upstream source for Germany (to isolate from other changes).

This cherry-pick's @ondenman’s remappings from #16622
but it looks like there are still some others to remap as well.


tmtmtmtm:
Woot!

PR #18305, opened 2016-10-07
Add rake task to remap ids in idmap files from memberships

octopusinvitro:
If the id of a politician changes, or a politician is given a UUID
because they didn't have one, we need to update our idmap files
to point the new ID at our UUID.

This adds a rake task to do that. You call it as:

```rake
be rake memberships:remap_ids[oldID,newID]
```

(some shells, e.g. zsh, may need to quote those square brackets)

P.S. Not sure if the `remap` helper method should be moved to the `UuidMapFile` class.

### DIFF
--- a/lib/uuid_map.rb
+++ b/lib/uuid_map.rb
@@ -44,5 +44,6 @@ class UuidMapFile
 
   def check_ids(data, from, to)
     abort "No existing data for #{from}" unless data[from]
+    abort "Already have data for #{to}" if data[to]
   end
 end
--- a/lib/uuid_map.rb
+++ b/lib/uuid_map.rb
@@ -38,6 +38,8 @@ class UuidMapFile
   def remap(from, to)
     @data ||= mapping
     check_ids(@data, from, to)
+    @data[to] = @data.delete(from)
+    rewrite(@data)
   end
 
   private
--- a/lib/uuid_map.rb
+++ b/lib/uuid_map.rb
@@ -34,4 +34,15 @@ class UuidMapFile
   def source
     [@newfile, @oldfile].find(&:exist?)
   end
+
+  def remap(from, to)
+    @data ||= mapping
+    check_ids(@data, from, to)
+  end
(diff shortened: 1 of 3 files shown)


octopusinvitro:
@tmtmtmtm I moved the helper method to the UUID mapper class

lib/uuid_map.rb
```diff
@@ -34,4 +34,18 @@ def rewrite(data)
...
+    @data[to] = @data.delete(from)
+    rewrite(@data)
+  end
+
+  private
+
+  def check_ids(data, from, to)
+    abort "No existing data for #{from}" unless data[from]
+    abort "Already have data for #{to}" if data[to]
+  end
```
tmtmtmtm: This is a slightly odd helper function on several fronts:
* It has very different behaviour from what I'd expect a method called `check_ids` to do
* it's divorced from the actual object
* it's very tightly coupled to one particular usage pattern (ie it does separate things that aren't necessarily inter-dependent).
* `abort`-ing out of it is very aggressive (and as you've seen, causes problems with testing)

I think this would be a lot better as an actual method (i.e. that doesn't take `data` as an argument), and probably actually _two_ quite simple query methods: one for checking if a given `id` exists in the file, and one for checking if a given `uuid` exists in the file.

lib/uuid_map.rb
```diff
@@ -34,4 +34,18 @@ def rewrite(data)
   def source
     [@newfile, @oldfile].find(&:exist?)
   end
+
+  def remap(from, to)
+    @data ||= mapping
```
tmtmtmtm: I'm not following why you're creating a new instance variable for this, especially as it doesn't seem to be used from any other method. Is there some reason why that's necessary/useful, or should this just be a local variable to this method?

octopusinvitro:
Rebased and redone in #18816

octopusinvitro:
Rebased and redone in #18816

PR #18449, opened 2016-10-11
Fix rake countries.json with no arguments

chrismytton:
This fixes `rake countries.json` so that it rebuilds all countries when there is no EP_COUNTRY_REFRESH environment variable set.

Fixes https://github.com/everypolitician/everypolitician/issues/522

### DIFF
--- a/Rakefile.rb
+++ b/Rakefile.rb
@@ -27,8 +27,12 @@ task 'countries.json' do
   # we only build any country that contains that string. For example:
   #    EP_COUNTRY_REFRESH=Latvia be rake countries.json
   to_build = ENV['EP_COUNTRY_REFRESH'] || 'data'
-  countries = EveryPolitician.countries.select do |c|
-    c.slug.downcase.include? to_build.downcase
+  if to_build == 'data'
+    countries = EveryPolitician.countries
+  else
+    countries = countries.select do |c|
+      c.slug.downcase.include? to_build.downcase
+    end
   end
 
   data = json_load('countries.json') rescue {}


chrismytton:
@tmtmtmtm :eyes:

Rakefile.rb
```diff
@@ -27,8 +27,12 @@ def json_write(file, json)
   # we only build any country that contains that string. For example:
   #    EP_COUNTRY_REFRESH=Latvia be rake countries.json
   to_build = ENV['EP_COUNTRY_REFRESH'] || 'data'
-  countries = EveryPolitician.countries.select do |c|
-    c.slug.downcase.include? to_build.downcase
+  if to_build == 'data'
+    countries = EveryPolitician.countries
```
tmtmtmtm: It's not clear from the Issue what was actually going wrong before, but this doesn't seem like the right fix, in that it's adding even more complexity, when I suspect the solution is actually to simplify, by removing something that's doing the wrong thing.

The use of `data` as the RHS of the `||` is so that we can fall through and get the right information later — it's a meaningful value, whereas this approach effectively is trying to make it meaningless in some contexts and not in others (i.e. this part would be much clearer with removing the `||` and dealing with nil/empty here instead). 

Is the problem that it's correctly generating the `commit_metadata` both ways, but then iterating over the wrong list of countries to update with that data?

tmtmtmtm: @chrismytton I decided to take this even further, and extract all this logic out into a Task class instead, which made it significantly easier to refactor: https://github.com/everypolitician/everypolitician-data/pull/18484

chrismytton: @tmtmtmtm Ah yep, that's a much nicer approach 👍

chrismytton:
Closing in favor of https://github.com/everypolitician/everypolitician-data/pull/18484

PR #18450, opened 2016-10-11
Refresh countries.json for all countries

chrismytton:
This does a full refresh of `countries.json` to catch a few countries where things had slipped a bit out of sync. Specifically this fixes Malaysia, which is currently pointing to a SHA that doesn't exist on `master`.

This uses the code from https://github.com/everypolitician/everypolitician-data/pull/18449 to perform the full rebuild.

### DIFF
--- a/countries.json
+++ b/countries.json
@@ -2160,7 +2160,7 @@
     "name": "Congo-Kinshasa (DRC)",
     "country": "Congo-Kinshasa (DRC)",
     "code": "CD",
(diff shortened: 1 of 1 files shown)


chrismytton:
Once this goes through it should fix [the broken viewer-sinatra builds on Travis](https://travis-ci.org/everypolitician/viewer-sinatra/builds/166796188#L2109).

PR #18495, opened 2016-10-12
Use the bundler github shortcut in Gemfile

octopusinvitro:
# What does this do?

Uses the `github:` shortcut in the `Gemfile` for gems that are not in Rubygems and we have to link to their repos.

# Why was this needed?

For security reasons as described in http://bundler.io/git.html#security

# Relevant Issue(s)

Part of https://github.com/everypolitician/everypolitician/issues/518

# Implementation notes

* There was no rake task for bundle:audit so [I added the gem in this PR](https://github.com/everypolitician/everypolitician-data/pull/18496)

* After that I ran `bundle exec rake bundle:audit` and got:
```bash
:) be rake bundle:audit
Updating ruby-advisory-db ...
De https://github.com/rubysec/ruby-advisory-db
 * branch            master     -> FETCH_HEAD
Already up-to-date.
Updated ruby-advisory-db
ruby-advisory-db: 273 advisories
No vulnerabilities found
```

# Screenshots

None

# Notes to Reviewer

None

# Notes to Merger

None


chrismytton:
> You need to use the ref: argument in the Gemfile for this change because you're switching from git:// to https://. The ref option needs to be the SHA that's in the current Gemfile.lock on the master branch.

So for `csv_to_popolo` I would look in [`Gemfile.lock` on `master`](https://github.com/everypolitician/everypolitician-data/blob/f7a7eac97da158711e999eeea309063e448d4e7c/Gemfile.lock#L35) and copy the SHA from there into the `ref` option, the short SHA (first 7 chars) is fine.

```diff
diff --git a/Gemfile b/Gemfile
index 8035322..74a57b1 100644
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'json'
 gem 'nokogiri'
 gem 'pry'
 gem 'rake'
-gem 'csv_to_popolo', '~> 0.27.0', github: 'tmtmtmtm/csv_to_popolo'
+gem 'csv_to_popolo', '~> 0.27.0', github: 'tmtmtmtm/csv_to_popolo', ref: 'ec39a8b'
 gem 'colorize'
 gem 'fuzzy_match'
 gem 'yajl-ruby', require: 'yajl'
```

Hopefully that makes sense!

tmtmtmtm:
Assigning back to @octopusinvitro because travis is failing.

Also, per the discussion on Slack about this, I think we should only point at a specific `ref:` where we really need to be at that precise point rather than up to date (and the reason for that should then be documented in the Gemfile). Where we end up switching protocol we should then effectively treat the task as 'update to latest version of gem' rather than jumping through hoops to pin unnecessarily to whatever we just happened to get last time we updated.

octopusinvitro:
Redid this in  #18542

octopusinvitro:
Redid this in #18542

PR #18590, opened 2016-10-14
Sierra Leona: update instrucions

octopusinvitro:
The scraper for the main source for Sierra Leona was moved in morph from Tony's account to EP-scrapers account. This PR updates that.

The new data coming after the update in the scraper fixes the image urls that were escaped twice.


tmtmtmtm:
Woohoo!

Issue #18489, opened 2016-10-12
Guernsey 2016 term needs adding


ondenman:
## Legislature

Guernsey/States

## Source

https://gov.gg/contactus

## Type of Data

Members

## Notes

We only have the 2012 term. The island had a general election in 2016: https://en.wikipedia.org/wiki/Guernsey_general_election,_2016

----

ondenman:
I've added a new scraper (https://morph.io/everypolitician-scrapers/guernsey-2016) but data isn't being written to morph. Seems like a problem their end (https://help.morph.io/t/nothing-saving-to-database/251/30)

PR #18817, opened 2016-10-17
Guernsey: Move 2012 membership data to manual file

ondenman:
The source website has changed to reflect the new 2016 term. As a result, the 2012 data is no longer online. To address this, the 2012 scraper data has been moved to a manual file.

This is a step towards fixing: https://github.com/everypolitician/everypolitician-data/issues/18489


tmtmtmtm:
👍

PR #18821, opened 2016-10-17
Guernsey: Swap name order

ondenman:
In preparation of adding term 2016 data -- where the names are formatted:
'first_name last_name' order -- this PR swaps the ordering of the 2012
term names.

This is a step towards fixing: https://github.com/everypolitician/everypolitician-data/issues/18489


tmtmtmtm:
👍

PR #18966, opened 2016-10-20
adds type to legislature metadata

struan:
fixes https://github.com/everypolitician/everypolitician/issues/526

### DIFF
--- a/lib/country_data.rb
+++ b/lib/country_data.rb
@@ -138,7 +138,8 @@ module Everypolitician
       end
 
       def type
-        legislature[:type] || ''
+        raise "Missing type for Legislature #{legislature[:name]} in #{dir}" unless legislature[:type]
+        legislature[:type]
       end
     end
   end
--- a/lib/country_data.rb
+++ b/lib/country_data.rb
@@ -103,12 +103,9 @@ module Everypolitician
       end
 
       def legislature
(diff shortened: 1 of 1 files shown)


lib/country_data.rb
```diff
@@ -128,6 +129,10 @@ def popolo
       def statement_count
         json_with_count.last
       end
+
+      def type
+        popolo[:organizations].find { |o| o[:classification] == 'legislature' }[:type] || ''
```
tmtmtmtm: I think it's probably worth removing the duplication with `lname` here by factoring out a `legislature` method that both call.

tmtmtmtm: The `|| ''` worries me a little here. I don't think we'd want an empty value ending up in `countries.json`. This data _should_ be available for every legislature (or at least will be when that country is being rebuilt, and thus its stanza here is being built), so I think I'd raise an exception in any case where this isn't filled in, so we can catch it early.

struan: Ok, there are currently 78 legislatures that don't have that data.

tmtmtmtm:
Thanks @struan — I'll try to make sure every legislature that's still missing these gets rebuilt over the weekend, so that the `everypolitician-ruby` code that consumes this can be ready to go early next week.

PR #18967, opened 2016-10-20
Fix REBUILD_SOURCE option

chrismytton:
This was broken when we switched to using a `Pathname` as we need to
call `#to_s` first, otherwise we get an error because there is no
`Pathname#include?` method.

Fixes https://github.com/everypolitician/everypolitician/issues/527


chrismytton:
@tmtmtmtm 👀

PR #18975, opened 2016-10-20
Rubocop: Ignore long blocks in tests

chrismytton:
Most of our tests are written "spec-style" and therefore have lots of long do...end blocks, this causes the block-length cop to complain, so disable that block for the tests.

Note that this includes `spec`, `t` and `test` because this base config is shared between all the EveryPolitician repos.

This will prevent failures like [this one](https://travis-ci.org/everypolitician/close_old_pull_requests/builds/169286902).


chrismytton:
@tmtmtmtm :eyes:

PR #19107, opened 2016-10-24
The Philipines: make manual file

octopusinvitro:
The members for the new term (17th Congress) are up in the official site, and the ones for term 16 were removed.

This PR moves the members in `morph/data.csv` to a manual file as a previous step before importing the new term.

When running `bundle exec rake clean default` nothing changed.


tmtmtmtm:
👍 

http://www.gov.ph/lists/current-congress-of-the-philippines/ still has the list of all these people, but it has less info, so isn't a suitable replacement source.

PR #19110, opened 2016-10-24
Hong Kong: Archive the 5th Legislative Council

mhl:
This adds a manual source for the 5th Legislative Council, and removes
fetching data from the scarper. The scraper will need minor changes, and
then a similar source can be reinstated to fetch the data for the 6th
Legislative Council.


data/Hong_Kong/Legislative_Council/sources/instructions.json
```diff
@@ -1,14 +1,9 @@
...
-      "file": "morph/data.csv",
-      "create": {
-        "from": "morph",
-        "scraper": "struan/hong_kong_legislative_council_members",
-        "query": "SELECT *, REPLACE(id,'yr12-16/','') AS id FROM data ORDER BY id"
-      },
+      "file": "manual/term-5.csv",
       "source": "http://www.legco.gov.hk/",
-      "type": "membership"
+      "type": "membership",
```
tmtmtmtm: This isn't valid JSON, so makes a build fail with

```
JSON::ParserError: parse error: invalid object key (must be a string)
             "type": "membership",     },     {       "file": "morph/w
                     (right here) ------^
```

(We've toyed with making this use a JSON5 parser just to avoid this entire class of problem, but haven't quite found the tuits yet)

PR #19111, opened 2016-10-24
Philippines: make terms manual

octopusinvitro:
We are moving the terms file to manual and will also update the scraper to stop recording the terms information.

Running `bundle exec rake clean default` produces no changes in the data.


tmtmtmtm:
👍

PR #19158, opened 2016-10-25
Hong Kong: maintain term metadata as a manual source

mhl:
This pull request switches to getting term metadata (e.g. start & end dates)
from a manually maintained terms.csv file in manual/sources. Previously
this data was essentially being manually maintained anyway, it's just that
that was happening in the source code of the scraper and being fetched
from the terms table that the scraper generated.

This also adds metadata for the 6th term.


mhl:
OK, I hope I've revised that as you want now - the fixups just looked confusing, so this is the rewritten branch. It somehow wasn't clear to me from what you said before that the scraper needn't be generating a terms table at all, although that seems clear looking back on it...

tmtmtmtm:
👍  — thanks @mhl

PR #19176, opened 2016-10-25
Philippines - House: Add term 17

octopusinvitro:
This PR adds data for the new term (term 17) in the Philippines

Checkbox list with links here:
https://gist.github.com/octopusinvitro/f748e068d2249978389bab1cb0edff18

Preserve old data *if necessary* (ep-data)
----------------------------------------

* [x] data no longer available upstream is in manual file <https://github.com/everypolitician/everypolitician-data/blob/master/data/Philippines/House/sources/manual/term-16.csv>
* [x] idmaps and reconciliation files preserved too <https://github.com/everypolitician/everypolitician-data/blob/master/data/Philippines/House/sources/idmap/term-16.csv>
* [x] redundant scraper no longer used in `instructions.json` <https://github.com/everypolitician/everypolitician-data/blob/master/data/Philippines/House/sources/instructions.json#L3-L7>
* [x] clean build works with no changes <https://github.com/everypolitician/everypolitician-data/pull/19107#issuecomment-255760085>

New term (ep-data)
------------------

These are in the PR to add the new data

* [x] a single `terms.csv` exists and has previous terms *and* this new term in 
* [x] previous term end date is correct, and source credited in commit message
* [x] new term start date is correct, and source credited in commit message

New data (ep-data)
------------------

* [x] new memberships are merged or reconciled
* [x] party groupings are correct (screenshot) --> I left a screenshot but I am not sure about 1-CARE and SAGIP

Reconciliations against other sources
------------------------------------

* [x] wikidata is being reconciled (or else an issue exists for it)

To run the site locally with these changes: 
```
DATASOURCE='https://cdn.rawgit.com/everypolitician/everypolitician-data/d11d067833c1c285c6f3832791f0314e9c634e1f/countries.json' foreman start
```

## New term
terms

## Party groupings
parties

PR #19223, opened 2016-10-26
Allow Merging of Membership sources with same IDs

tmtmtmtm:
Previously we haven't needed to merge Membership sources with the same ID, as those will almost always be coming from the same upstream source, and will therefore be in the same Morph database, and can be fetched together as a single 'source'.

Sometimes, however, we will have archived off a previous term to a manual source file, and will want to merge this with a new source file for the latest term from the same upstream source, with shared IDs.

In such a case — i.e. where we have an `instructions.json` stanza for a second Membership source, and no `merge` details, we can simply re-use any ID → UUID mapping we already know of for any of the people in this source.

## Notes to Reviewer

There are no direct tests for this functionality, but it's easy to test the behaviour exhaustively by running a build of all the existing data. With this change, every existing legislature still builds without material change (we still don't have a 100% deterministic process).

NB: This was added for https://github.com/everypolitician/everypolitician-data/pull/19176, but it turns out that that source only _appears_ to share IDs: ferrer-j in the 16th term was Jeffrey P. Ferrer, but is Juliet Marie D. Ferrer in the 17th!

### DIFF
--- a/lib/source/membership.rb
+++ b/lib/source/membership.rb
@@ -29,6 +29,9 @@ module Source
 
         pr = reconciler.reconciliation_data rescue abort($!.to_s)
         pr.each { |r| id_map[r[:id]] = r[:uuid] }
+      else
+        # reuse any IDs we already have from other sources
+        csv.each { |r| id_map[r[:id]] &&= r[:uuid] }
       end
 
       # Generate UUIDs for any people we don't already know


lib/source/membership.rb
```diff
@@ -29,6 +29,9 @@ def merged_with(csv)
 
         pr = reconciler.reconciliation_data rescue abort($!.to_s)
         pr.each { |r| id_map[r[:id]] = r[:uuid] }
+      else
```
tmtmtmtm: This `else` bugs me a little (as always). It would be good to split this out into a series of different strategies, each with much simpler logic, but that seems like too big a change for right now. I don't think this makes the existing code significantly more complex (if anything there's now a nice symmetry between lines 31 and 34) where previously there was the dangling question of "but what if there are no `merge_instructions`), so I think we can live with it for a while whilst the whole build process is in flux anyway. (This whole thing is going to need to be completely changed once we start adding `Person`s first, and then building up their `Membership`s afterwards, rather than vice versa)

tmtmtmtm:
@chrismytton 👀

lib/source/membership.rb
chrismytton: > This else bugs me a little (as always).

Interesting! Is that just because `else` statements increase the complexity of the method? Or is it a more philosophical objection?

tmtmtmtm: Philosophical. I'm (rather obviously) not a complete purist on this, but an `else` in OO code is almost always always a very bad smell. The OO way is to replace pretty much all conditionals with polymorphism. My standard rule of thumb is to treat as very suspicious any `if` that isn't either a guard clause or part of a factory method (and even in the latter case, data-driven is usually better than a giant case statement)

PR #19240, opened 2016-10-26
Hong Kong: add memberships of the 6th term from morph

mhl:
```
Add memberships from sources/manual/term-5.csv
Add memberships from sources/morph/term-6.csv
Merging with sources/morph/wikidata.csv
  ⚥ data for 0; 0 added


Top identifiers:
  98 x wikidata
  17 x freebase
  5 x viaf
  2 x lcauth
  1 x fast

Creating names.csv
Creating unstable/positions.csv
Persons matched to Wikidata: 98 ✓ 
Parties matched to Wikidata: 21 ✓ 
```


data/Hong_Kong/Legislative_Council/sources/instructions.json
```diff
@@ -6,6 +6,21 @@
...
+        "scraper": "everypolitician-scrapers/hong_kong_legislative_council_members",
+        "query": "SELECT *, REPLACE(id,'yr16-20/','') AS id FROM data ORDER BY id"
+      },
+      "source": "http://www.legco.gov.hk/",
+      "type": "membership",
+      "merge": {
+        "incoming_field": "name",
+        "existing_field": "name",
+        "reconciliation_file": "reconciliation/term-6.csv"
+      }
```
tmtmtmtm: The IDs look as if they're consistent between terms, which means you shouldn't need this merge block (or the reconciliation file).

mhl:
OK, hopefully I've added those Wikidata parties correctly, and I've added a commit with the update to `countries.json`.

tmtmtmtm:
Staged:

screen shot 2016-10-27 at 11 25 13

Wikipedia:

screen shot 2016-10-27 at 11 26 25

mhl:
@tmtmtmtm Oops, sorry, that was stupid (messing up the header row). I've made a PR to fix that here: https://github.com/everypolitician/everypolitician-data/pull/19308

If that's OK, then I'll rebase this over master and make the "Unknown" -> "Independent" change in this PR.

mhl:
(Sorry about the delay - sysadmin duties then lunch.) Here's a PR for the scraper update: https://github.com/everypolitician-scrapers/hong_kong_legislative_council_members/pull/2

I made the change there rather than in the instructions.json's query since:

```
SELECT id, term, name, honorific_suffix, honorific_prefix, img, (CASE WHEN faction = '' THEN 'Independent' ELSE faction END) as faction, email, website, phone, fax, gender, source, area_type, area FROM data ORDER BY id;
```

... would be annoying to maintain, e.g. we'd need to manually add any new columns, and if I did:

```
SELECT (CASE WHEN faction = '' THEN 'Independent' ELSE faction END) as faction, * FROM data ORDER BY id;
```

... we'd be relying (if I've followed through the code right) on Ruby's `CVS.table`'s handling of multiple columns with the same header name.

tmtmtmtm:
@mhl I'm pretty sure morph's implementation doesn't allow for multiple "columns" with the same name in a SELECT — i.e. "SELECT *, (…) AS faction" will happily override the version from within the `*`.

So the example would need to be:

```sql
  SELECT *, (CASE WHEN faction = '' THEN 'Independent' ELSE faction END) AS faction
    FROM data 
   ORDER BY id
```

It is, of course, possible that that is unstable behaviour, so setting it upstream seems sensible.

mhl:
Trying viewer-sinatra with `DATASOURCE='https://raw.githubusercontent.com/everypolitician/everypolitician-data/8489ce05a1294bdac75e9998ecbd50235a48335c/countries.json'` they do now appear as "Independent":

hong-kong-independents

mhl:
I realise now that it would have been a good idea to do the `s/unknown/Independent/` change for the old 5th term too, but I'm keen to just get the merged soon, so maybe that can be done afterwards.

tmtmtmtm:
@mhl 👍 , though it looks like you forgot to rebuild the Group data after adding the WIkidata mapping for Independents, so I added another commit to do that.

It's fine not to change the handing of Independents in the prior term here, but it would be good if you could open an Issue to remind us.

PR #19460, opened 2016-10-31
Check source CSVs for well-formedness

tmtmtmtm:
Use the ODI csv linter to make sure that our source CSVs (particularly the ones  we generate ourselves, such as terms files, reconciliation files etc, are well  formed before attempting a build.

This can be run manually from within a legislature sub-directory with `rake csvlint:validate`, but it is also set to be a default rake task, and thus will happen automatically in-between the `clean` or `clobber` and the start of the generation process in our normal approach.

Closes https://github.com/everypolitician/everypolitician/issues/534

## Notes to reviewer

1. This gem unfortunately requires us to up our minimally-supported ruby version. If there's a way to conditionally run these checks only if a suitable ruby version is installed, I'd be quite tempted to do that instead, but it seems to be a non-trivial amount of effort for code that is very rarely run by anyone outside the team (and even then by people who are unlikely to have a problem with using 2.3). This approach also lets us evolve towards a much more complex set of tests using JSON schemas.
1. This can be seen failing against the current `master` version of The Philippines. Many other problems were fixed separately in https://github.com/everypolitician/everypolitician-data/pull/19488
1. This currently fails if a CSV flle contains linebreaks within a field (e.g. with Kenya). I'm currently undecided whether we want to override that or not. In theory it should be perfectly acceptable to have these, but in practice we're almost never going to want them, and should be cleaning them up, either in the upstream scraper, or at input. So I'm currently leaning towards using this to force us to resolve them. (e.g. the Kenya issue is from https://github.com/mysociety/pombola/issues/2160 which we really should fix)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-abort 'Ruby should be >= 2.1.0' unless RUBY_VERSION.to_f >= 2.1
+abort 'Ruby should be >= 2.3' unless RUBY_VERSION.to_f >= 2.3
 git_source(:github) { |repo_name| "https://github.com/#{repo_name}.git" }
 
 gem 'json'
@@ -34,4 +34,5 @@ group :test do
   gem 'webmock'
   gem 'rubocop'
   gem 'flog'
+  gem 'csvlint'
 end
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
@chrismytton 👀

tmtmtmtm:
@octopusinvitro this should act as a suitable regression test for your Philippines file.

PR #19623, opened 2016-11-02
Slovakia: Move terms file to manual

ondenman:
Term information was being sourced from http://api.parldata.eu
We're moving the terms to a manual file, since this source
is no longer being maintained.

This is a step towards adding a new term.

### DIFF
--- a/countries.json
+++ b/countries.json
@@ -8893,11 +8893,11 @@
         "slug": "National-Council",
         "sources_directory": "data/Slovakia/National_Council/sources",
         "popolo": "data/Slovakia/National_Council/ep-popolo-v1.0.json",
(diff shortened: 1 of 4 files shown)


data/Slovakia/National_Council/unstable/stats.json
```diff
@@ -9,7 +9,7 @@
   },
   "terms": {
     "count": 5,
-    "latest": "2012-03-11"
+    "latest": ""
```
tmtmtmtm: We can't let this go through with this. It turns out that this is a wider problem — https://github.com/everypolitician/everypolitician/issues/537. But this is going to have to wait on that being fixed first.

tmtmtmtm:
I rebased this to get the benefit of https://github.com/everypolitician/everypolitician-data/pull/19626

Issue #19644, opened 2016-11-02
Slovakia: Reconcile new member data with wikidata


ondenman:
## Legislature

Slovakia National Council

## Source

https://en.wikipedia.org/wiki/List_of_members_of_the_National_Council_of_Slovakia,_2016–

## Type of Data

Wikidata

## Notes

The legislature is being updated to add the latest term. The wikidata scraper needs to be updated to scrape wikidata for the members of this term.

PR #19649, opened 2016-11-02
Reconcile new member wikidata

ondenman:
Reconciles wikidata for 2016 members.

Closes: https://github.com/everypolitician/everypolitician-data/issues/19644


tmtmtmtm:
@ondenman how convinced are you that the Tibor Jančula here is the same person as the footballer? I suspect he _probably_ is, as the date of birth is the same, but I haven't yet been able to find anything to confirm that.

tmtmtmtm:
General consensus is that they're same person, so good enough for me!

PR #19752, opened 2016-11-03
Laos Assembly: add 2016 term

ondenman:
This PR imports the latest term.

Both 2011 and 2016 terms are pulled in from the same source but as we have no upstream IDs and members' names are spelled differently across terms we need to reconcile the memberships manually.

149 in total:
screen shot 2016-11-03 at 11 16 35

screen shot 2016-11-03 at 13 43 31

Sanity check
------------

* [x] election has happened and new term is not in EveryPolitician already
* [x] you're looking at the *correct* house for *correct* country
* [x] there's not a much more urgent legislature you could be doing, is there?

New term (ep-data)
------------------

* [x] a single `terms.csv` exists and has previous terms *and* this new term in it
* [x] previous term end date is correct, and source credited in commit message
* [x] new term start date is correct, and source credited in commit message

New data (ep-data)
------------------

* [x] new memberships are merged or reconciled
* [x] party groupings are correct (screenshot) (party data not available from source site)

Reconciliations against other sources
------------------------------------

* [x] wikidata is being reconciled (or else an issue exists for it)

Pull request
------------

* [x] has country: legislature name in title
* [x] has screenshot of staged PR-site (especially "party groupings")
* [x] Travis tests pass
* [x] is assigned to reviewer


tmtmtmtm:
@ondenman this mostly looks fine, but it's a little hard to reason about as it's doing two different things with different behaviour (renaming the old file, which should have no changes) and adding new data (with lots of changes). Could you split the rename out to a separate PR?

ondenman:
Closing in favour of breaking into separate PRs:

- Step 1: Move the 2011 data to separate file (https://github.com/everypolitician/everypolitician-data/pull/20878)

PR #20147, opened 2016-11-07
Ensure "Unknown" party has an ID

tmtmtmtm:
Often upstream sources don't provide us any party/group information for Members. Our practice is to default such cases to Unknown unless we explicitly know that they are Independent.

When doing that, we usually construct a pseudo-Party named "Unknown". However, if the source explicitly sets the party field to "Unknown", we use the Popolo Organization generated from that (in CSV-to-Popoolo) instead.

When we generate such a Party ourselves, we ensure it has an ID ("party/_unknown"). However, when we re-use an existing "Unknown" party, we need to also ensure that it has an ID too, otherwise we end up with lots of broken references.

Part of https://github.com/everypolitician/everypolitician/issues/536 (though Greece and Sweden are both still failing for other, though related, reasons)

## Note to Reviewer

This is a slightly unsatisfying fix to this. It's adding complexity to this process, rather than removing it. However, we're hoping to completely rework the whole way Parties and their IDs work RSN anyway (https://github.com/everypolitician/everypolitician/issues/428), and this fixes some failing builds, so I think we have to live with it for a while.


tmtmtmtm:
@chrismytton 👀

tmtmtmtm:
It would be quite nice if the review bot could report better on Membership with field changes, rather than reporting them all as removals and additions…

PR #19629, opened 2016-11-02
Slovakia National Council: Add 2016 term

ondenman:
Adds new term from new scraper source.

The latest term has not been added to the source from which the previous terms have been scraped. As such, a new scraper is required. This PR adds the new scraper.

screen shot 2016-11-02 at 12 12 03

screen shot 2016-11-02 at 12 15 13

Sanity check
------------

* [x] election has happened and new term is not in EveryPolitician already
* [x] you're looking at the *correct* house for *correct* country
* [x] there's not a much more urgent legislature you could be doing, is there?

New term (ep-data)
------------------

* [x] a single `terms.csv` exists and has previous terms *and* this new term in it
* [x] previous term end date is correct, and source credited in commit message
* [x] new term start date is correct, and source credited in commit message

New data (ep-data)
------------------

* [x] new memberships are merged or reconciled
* [x] party groupings are correct (screenshot)

Reconciliations against other sources
------------------------------------

* [x] wikidata is being reconciled (or else an issue exists for it) (Wikidata will be reconciled in a later commit. See issue: https://github.com/everypolitician/everypolitician-data/issues/19644)

Pull request
------------

* [x] has country: legislature name in title
* [x] has screenshot of staged PR-site (especially "party groupings")
* [x] Travis tests pass
* [x] is assigned to reviewer

PR #20671, opened 2016-11-18
Slovakia (National Assembly): Fix source url

chrismytton:
There was an unwanted `'` character at the beginning of the source url in the Slovakia `instructions.json` which was causing an invalid link in the viewer-sinatra HTML, which in turn prevented http://everypolitician.org/ from being deployed correctly. This invalid url was originally introduced in https://github.com/everypolitician/everypolitician-data/pull/19629.

This change just fixes the source url, I'll open a separate pull request to make some changes to ensure we catch this in the future.

### DIFF
--- a/countries.json
+++ b/countries.json
@@ -8895,11 +8895,11 @@
         "slug": "National-Council",
         "sources_directory": "data/Slovakia/National_Council/sources",
         "popolo": "data/Slovakia/National_Council/ep-popolo-v1.0.json",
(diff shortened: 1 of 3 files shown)


chrismytton:
Pull request to catch and prevent this in the future is here https://github.com/everypolitician/everypolitician-data/pull/20672.

tmtmtmtm:
cc @ondenman

PR #20672, opened 2016-11-18
Ensure all source urls are valid

chrismytton:
Following on from https://github.com/everypolitician/everypolitician-data/pull/20671 this adds some extra logic to ensure the source urls in `instructions.json` are actually valid. Because we're running the urls though `URI.encode(URI.decode(url))` we're now getting the "correct" but ugly version of the url appearing in the Popolo JSON, which doesn't seem ideal but I can't immediately think of a way to avoid.


tmtmtmtm:
@chrismytton have you had any further thoughts on this in the meantime? I suspect it's probably OK for the Popolo to always contain 'correct' URLs, though it would be good to re-prettify those again on everypolitician.org

PR #20883, opened 2016-11-21
Laos Assembly -- new term 2016

ondenman:
This PR imports the latest term.

Both 2011 and 2016 terms are pulled in from the same source but as we have no upstream IDs and members' names are spelled differently across terms we need to reconcile the memberships manually.

149 in total:
screen shot 2016-11-03 at 11 16 35

screen shot 2016-11-21 at 13 16 04


tmtmtmtm:
👍

PR #20944, opened 2016-11-22
Hopefully fix intermittent UUID test error

chrismytton:
This is a hard to debug failure, but I think this test is occasionally
failing because we're not holding onto the reference to the tempfile
that we're using, so it's getting garbage collected before we can read
from it.

This should fix errors like https://travis-ci.org/everypolitician/everypolitician-data/builds/177861445.

This follows on from https://github.com/everypolitician/everypolitician-data/pull/20899


chrismytton:
@tmtmtmtm 👀

PR #21278, opened 2016-11-29
Argentina Senate add 2016 term

ondenman:
The Argentina Senate has no set terms. We are now considering a term to run from the 1st of March to the 30th of November each year.

The following has been added to the Senate scraper. (https://github.com/everypolitician-scrapers/argentina-senate/commit/a93b937ac40714806062e172986567091a529464). I have changed the 2015 term dates accordingly.
~~~ 
# The terms of senators are 6 years long, but offset by two or four
# years from their colleagues, so we're assuming that each year is a
# different term. Each session runs from the 1st of March to the 30th
# of November each year, [1] so treat those as the start and end dates
# of each "term".
# [1] https://en.wikipedia.org/wiki/National_Congress_of_Argentina
terms = (2015..Date.today.year).map do |year|
  {
    id: year,
    start_date: Date.new(year, 3, 1).to_s,
    end_date: Date.new(year, 11, 30).to_s,
  }
end
~~~

## Note to reviewer

I'm not too sure how helpful the Wikipedia data is in this case. It is sourced from [this](http://www.senado.gov.ar/parlamentario/bloques/) site which is currentling 500ing. Wikipedia's page [List of current Argentine Senators](https://en.wikipedia.org/wiki/List_of_current_Argentine_senators) states the list is correct as of March 2010.

There is a [list of senators](http://www.senado.gov.ar/senadores/listados/agrupados-por-bloques) on the .gov.ar source site.

screen shot 2016-12-05 at 10 21 37
screen shot 2016-12-05 at 11 05 46

This changes the existing dates for these new dates.

* [x] new term added to [[terms.csv]]
* [x] new term start date is correct, and source credited in commit message
* [x] previous term end date is set (and, if added here, source is credited in commit message)
* [x] members share IDs with previous terms, or are reconciled
* [x] number of members in the term seems correct
* [x] Wikidata scraper updated (if appropriate), and members reconciled (nothing to reconcile)
* [x] new parties/groups added to [[group_wikidata.csv]] and rebuilt
* [x] countries.json updated
* [x] PR information set
    * [x] PR title includes country + legislature name
    * [x] PR is tagged "New term"
    * [x] Screenshots added of party groupings on staged EP.org vs Wikipedia
    * [x] "Notes to Reviewer" includes information on any non-trivial Reconciliation decisions
    * [x] If any scrapers have moved, "Notes to Merger" includes reminder to delete the old scraper


ondenman:
@tmtmtmtm There is still no 2016 term but I assume that this will be handled in wikidata now?

PR #21284, opened 2016-11-29
WIP -- Argentina senate change 2015 term dates

ondenman:
The Argentina Senate has no set terms. We are now considering a term to run from the 1st of March to the 30th of November each year.

The following has been added to the Senate scraper. (https://github.com/everypolitician-scrapers/argentina-senate/commit/a93b937ac40714806062e172986567091a529464). I have changed the 2015 term dates accordingly.
~~~ 
# The terms of senators are 6 years long, but offset by two or four
# years from their colleagues, so we're assuming that each year is a
# different term. Each session runs from the 1st of March to the 30th
# of November each year, [1] so treat those as the start and end dates
# of each "term".
# [1] https://en.wikipedia.org/wiki/National_Congress_of_Argentina
terms = (2015..Date.today.year).map do |year|
  {
    id: year,
    start_date: Date.new(year, 3, 1).to_s,
    end_date: Date.new(year, 11, 30).to_s,
  }
end
~~~

This changes the existing dates for these new dates.


ondenman:
Closing in favour of: https://github.com/everypolitician/everypolitician-data/pull/21278

PR #22284, opened 2016-12-12
Kosovo (Assembly) — rebuild with moved Sources

tmtmtmtm:
Rebuild data with "source" information moved from Person to Membership


tmtmtmtm:
issue with unprintable characters in names opened as https://github.com/everypolitician/everypolitician/issues/545

PR #22971, opened 2016-12-21
Increase our default allowable LineLength to 120

tmtmtmtm:
In the rubocop configuration we inhert from in pretty much every other
repo, set the default max LineLength to 120 characters. Individual repos
can still increase, or decrease, this at will, but this will give them a
little more latitude to start with.


tmtmtmtm:
Merging for now. Can revisit when @chrismytton returns after Christmas.

PR #22980, opened 2016-12-22
Slovenia (National Assembly): switch Morph source to EP-scrapers

tmtmtmtm:
This was still pointing at the version in @chrismytton's own account

(Chris: you should also now delete https://morph.io/chrismytton/slovenia-national-assembly-wikidata)


chrismytton:
> (Chris: you should also now delete https://morph.io/chrismytton/slovenia-national-assembly-wikidata)

Thanks, done!

Issue #23810, opened 2017-01-04
Assemblée Nationale -- incorrect term name


ondenman:
We currently have a single term for the legislature. The terms file looks like this:

```
id,name,start_date
6,6e législature,2015-04-26
```

## The Problem

It looks like we've given the current term the wrong name and ID. The sixth term began in 2011 (https://fr.wikipedia.org/wiki/Assemblée_nationale_(Bénin)). The current term is actually the 7th.

## Proposed Solution

Amend the terms file.

PR #23812, opened 2017-01-04
Benin National Assembly -- amend term data

ondenman:
The election year for term 6 was in 2011 (https://fr.wikipedia.org/wiki/Assemblée_nationale_(Bénin)).

The current term is actually the 7th. (https://translate.google.co.uk/translate?hl=en&sl=fr&u=http://www.les4verites.info/Les-83-probables-deputes-elus-pour.html&prev=search)

This PR changes the term name only.

Closes: https://github.com/everypolitician/everypolitician-data/issues/23810

screen shot 2017-01-04 at 17 32 38


tmtmtmtm:
(Assuming there's another change coming that will fix the term ID upstream too…)

Issue #21869, opened 2016-12-08
Nigeria - current data is actually House of Representatives


octopusinvitro:
# Problem with Existing Data

In Nigeria, the Senate and the House of Representatives are combined in a bicameral body called [the National Assembly](https://www.wikidata.org/wiki/Q2635070).

![wikidata](https://cloud.githubusercontent.com/assets/2157089/21007752/2b3a42c2-bd36-11e6-875a-9b7361d4785c.png)

The data we actually have for Nigeria is labelled as "National Assembly". However, since the Senate has 109 members and the House of Representatives has 360, it looks like the data we have there is really for the House of Representatives.

![nigeria](https://cloud.githubusercontent.com/assets/2157089/21007779/4d147070-bd36-11e6-8b7f-9ebdb217ff5f.png)

## Legislature

Nigeria - National Assembly

## Problem

The data we have seems to be for the House of Representatives only, however, it appears as "National Assembly".

## Proposed Solution

Rename it to House of Representatives

## Notes

As possible sources for scraping:
The official site seems to separate the Senate http://www.nass.gov.ng/mp/senate
and the House of Representatives http://www.nass.gov.ng/mp/house

PR #24138, opened 2017-01-11
Nigeria: rename Lower House to House of Representatives

tmtmtmtm:
The National Assembly is the parliament as a whole (including the Senate).

screen shot 2017-01-11 at 11 40 41

Closes https://github.com/everypolitician/everypolitician-data/issues/21869


tmtmtmtm:
The travis-ci/pr failure is because the summarizer can't compare the two versions of the Popolo, as they're no longer in the same place. (We probably could make it cope with that, but I think this is such an edge case that it's not worth it.)

tmtmtmtm:
@mhl / @octopusinvitro not sure if this will have any direct, impact on what you're doing at the moment, but I'll hold off on merging until you give the go-ahead.

mhl:
@tmtmtmtm It won't create any problems for us if you merge this - thanks!

PR #24143, opened 2017-01-11
Benin Assembly --  update term id

ondenman:
The current term was incorrectly named the _6th_ when in fact it is the _7th_. The term name was updated in a previous PR. The scraper has since been updated to record the current term's ID as 7.

The scraper location will need to be moved to ep-scrapers. This PR updated the terms file with the correct term id and imports the term id as 7.


ondenman:
@tmtmtmtm I generated a new data file with `q` using a query that updated the term id. The old file was replaced with this new file.

tmtmtmtm:
@ondenman — this is a generated file; you can't just modify it. Your changes will be blown away on the next rebuild.

tmtmtmtm:
@ondenman what's the status of this PR?

ondenman:
Since this PR has become a little noisy, I'm closing it in favour of https://github.com/everypolitician/everypolitician-data/pull/43201

PR #24430, opened 2017-01-18
Syria Majlis -- add term dates

ondenman:
This is a step towards adding the 2016 term.

The end date of the 2012 term has been recorded as the day before the 2016 election.
The start date of the 2016 term has been recorded as the date of the first session of parliament.

Source: http://www.ipu.org/parline-e/reports/2307_E.htm

### DIFF
--- a/data/Syria/Majlis/sources/manual/terms.csv
+++ b/data/Syria/Majlis/sources/manual/terms.csv
@@ -1,2 +1,3 @@
-id,name,start_date,end_date,source
-2012,2012–,2012-05-24,,
+id,name,start_date,end_date,wikidata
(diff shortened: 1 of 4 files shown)


data/Syria/Majlis/sources/manual/terms.csv
```diff
@@ -1,2 +1,3 @@
-id,name,start_date,end_date,source
-2012,2012–,2012-05-24,,
+id,name,start_date,end_date,wikidata
+2012,2012–2016,2012-05-24,2016-04-12,Q28379948
+2016,2016-,2016-6-6,,Q28379971
```
tmtmtmtm: '2016-6-6' isn't a valid date format. This will need to be '2016-06-06'

PR #24436, opened 2017-01-18
Syria Majlia -- archive members

ondenman:
This is a step towards importing the 2016 term.

The source site has changed. Since we're flushing out the morph db on each run of the scraper, we're losing data. Some members are no longer listed on the site so we're archiving them off.


data/Syria/Majlis/ep-popolo-v1.0.json
```diff
@@ -516,7 +516,7 @@
       "name": "اسكندر جرادة"
     },
     {
-      "birth_date": "1937-01-01",
+      "birth_date": "2000-01-01",
```
tmtmtmtm: this seems like an unlikely change…

ondenman:
Closing in favour of https://github.com/everypolitician/everypolitician-data/pull/24709

PR #24448, opened 2017-01-18
Syria Majlis -- archive off term 2012 images

ondenman:
This is a step towards adding the new term (which also involves this next step).

We're no longer getting images from upstream since they're no longer on the sources pages and we're clearing out the Morph db on each run.

However, as the image URLs are still live, they can still be used.


data/Syria/Majlis/sources/archive/term-2012-images.csv
```diff
@@ -0,0 +1,275 @@
...
+323,http://parliament.gov.sy/SD08/msf/1429788914_.png
+324,http://parliament.gov.sy/SD08/msf/1421754634_.jpg
+325,http://parliament.gov.sy/SD08/msf/1421754919_.jpg
+326,http://parliament.gov.sy/SD08/msf/1421755207_.jpg
+327,http://parliament.gov.sy/SD08/msf/1429701135_.png
+330,
+331,
+332,http://parliament.gov.sy/SD08/msf/1434363648_.png
+363,
+364,
```
tmtmtmtm: It's a little odd to include all these blank lines here — that's essentially saying that there's explicitly no image for these people, rather than we just don't know of one.

PR #24455, opened 2017-01-18
Belgium (Chamber of Representatives): refresh data

everypoliticianbot:
Automated data refresh for Belgium - Chamber of Representatives

#### Output

```
Add memberships from sources/morph/wikipedia.csv
annick_lambrecht -> 7f8a65d8-7350-4059-bfbd-a2b13b243f4e
Merging with sources/morph/official.csv
Data Mismatches
* 17 of 169 unmatched
	{:id=>"01059", :name=>"Marie-Christine Marghem"}
	{:id=>"04666", :name=>"Theo Francken"}
	{:id=>"01079", :name=>"Melchior Wathelet"}
	{:id=>"01310", :name=>"Zakia Khattabi"}
	{:id=>"00795", :name=>"Pieter De Crem"}
	{:id=>"06850", :name=>"Koen Geens"}
	{:id=>"01238", :name=>"Jan Jambon"}
	{:id=>"00761", :name=>"André Flahaut"}
	{:id=>"00742", :name=>"Didier Reynders"}
	{:id=>"00657", :name=>"Johan Vande Lanotte"}
Merging with sources/morph/wikidata.csv
Data Mismatches
* 2 of 170 unmatched
	{:id=>"Q946183", :name=>"Johan Vande Lanotte"}
	{:id=>"Q17836281", :name=>"Annick Lambrecht"}
Adding GenderBalance results from sources/gender-balance/results.csv
  ⚥ data for 120; 0 added


Top identifiers:
  168 x wikidata
  37 x freebase
  27 x viaf
  23 x belgian_senate
  18 x flemish_parliament

Creating names.csv
Persons matched to Wikidata: 168 ✓ | 1 ✘
  No wikidata: Annick Lambrecht (7f8a65d8-7350-4059-bfbd-a2b13b243f4e)
Parties matched to Wikidata: 13 ✓ 

```

PR #24499, opened 2017-01-19
Belgium (Representatives): Reinstate missing wikipedia member

chrismytton:
This uses the output from https://github.com/everypolitician-scrapers/belgium-representatives-wikipedia/pull/1 to reinstate the member that went missing most recently in https://github.com/everypolitician/everypolitician-data/pull/24455.


chrismytton:
> looks like you forgot to regenerate countries.json

Oops. so I did! Pushed now.

tmtmtmtm:
The memberships are all changing due to https://github.com/everypolitician-scrapers/belgium-representatives-wikipedia/commit/841be960d373bd7c738be8e8ba8a1925006b122c, which is more accurate version of where the membership data is actually coming from.

PR #24608, opened 2017-01-24
Alderney States: Add new term dates

ondenman:
The legislature has an 'ordinary' election every two years where five
of the standing legislature stand for re-election. There are no formal
terms therefore I am creating a 'virtual' term. The start date is recorded
as the start date of members elected in 2016
(http://www.alderney.gov.gg/article/157566/Mike-Dean).

The end date is recorded as the end date of members elected in 2014
(http://www.alderney.gov.gg/article/4386/Steve-Roberts)

This is a step towards importing the new term data.


data/Alderney/States/sources/manual/terms.csv
```diff
@@ -1,2 +1,3 @@
 id,name,start_date,end_date,source
 2014,2014–2016,2014-11-22,2016-11,
+2017,2017-2018,2017-01-01,2018-12-31,http://www.alderney.gov.gg/article/4386/Steve-Roberts
```
tmtmtmtm: I don't think this is a particularly good `source` entry here, as it doesn't obviously say anything more broad than the single member. But generally we're moving away from having `source` entries in this file anyway, so I'd just drop the column. The citation in the commit message is good enough here if someone needs to trace the provenance of this.

ondenman:
@tmtmtmtm 👀

PR #24886, opened 2017-01-30
AI: remove spurious comma from instructions.json

davewhiteland:
Trailing comma was breaking JSON parsing, this fixes it.

This PR is specific to an error in Anguilla (AI)'s `instructions.json`

Thoughts:

A wee bit surprised a bad instructions file wasn't breaking things earlier, but on reflection it would only matter when upstream data triggers a rebuild, so such errors are effectively local to legislatures. Maybe worth adding JSON linting as well as the (broadly successful imo) CSV linting?

### DIFF
--- a/data/Anguilla/Assembly/sources/instructions.json
+++ b/data/Anguilla/Assembly/sources/instructions.json
@@ -12,7 +12,7 @@
     },
     {
       "file": "manual/terms.csv",
-      "type": "term",
+      "type": "term"
     },
     {
       "file": "wikidata/groups.json",


tmtmtmtm:
These files are processed as JSON5, not plain JSON, deliberately so that we don't need to care about this sort of issue.

Anything reading these files should be using the `Instructions` class, rather than rolling their own parsing. If there are places that aren't, then we should definitely be fixing those to do so.

davewhiteland:
@tmtmtmtm roger that, sorry for being a scaredy cat, but I'm older and wiser.

Background reading for purists who might stumble upon this PR: [Parsing JSON is a Minefield](http://seriot.ch/parsing_json.php)

PR #24900, opened 2017-01-30
Nigeria: add the Nigerian Senate

mhl:
n.b. there is a previous PR to add the Nigerian Senate based on
data from Pombola: https://github.com/everypolitician/everypolitician-data/pull/5440
This version is based on live data from the site so only has current, not historic,
members of the Senate.


data/Nigeria/Senate/sources/instructions.json
```diff
@@ -0,0 +1,34 @@
...
+        "from": "group-wikidata",
+        "source": "manual/group_wikidata.csv"
+      }
+    },
+    {
+      "file": "wikidata/elections.json",
+      "type": "wikidata-elections",
+      "create": {
+        "from": "election-wikidata",
+        "base": "Q22283753"
```
tmtmtmtm: I'm not sure that Q22283753 works as a base here. This includes _all_ general elections in Nigeria, even when those didn't include Senate elections. As far as I can tell, the Senate was only created in 1999, so we shouldn't really be importing elections from before then.

data/Nigeria/Senate/sources/manual/terms.csv
```diff
@@ -0,0 +1,2 @@
+id,name,start_date,end_date,source
+2015,2015–,2015-06-09,,https://en.wikipedia.org/wiki/National_Assembly_(Nigeria)
```
tmtmtmtm: Per https://github.com/everypolitician/everypolitician/wiki/terms.csv it would be better to have a `wikidata` column there than a `source`. [Q28555175](https://www.wikidata.org/wiki/Q28555175) is the 8th Assembly.

mhl: Thanks, I've changed that.

data/Nigeria/Senate/sources/instructions.json
mhl: I can't find a better Wikidata item for this (there doesn't seem to be one just for senatorial elections in Nigeria), so I've removed this source.

tmtmtmtm:
Data here:
screen shot 2017-02-06 at 09 51 15

https://en.wikipedia.org/w/index.php?title=Senate_of_Nigeria&oldid=759161215:
screen shot 2017-02-06 at 09 54 30

mhl:
> 1. It seems that Nigeria does have numbered terms, so it would be good to change this to be the 8th Assembly, rather than just the "2015–" term. However, it would be best to do that for the lower house as well, so I'll take your steer on whether to wait until this is through and then rename both, or whether to do a separate PR first for the House of Representatives, and then update this one in place here.

It would be great if we could leave that until after this is merged - I've created an issue for that in this repository, and assigned it to myself: https://github.com/everypolitician/everypolitician-data/issues/25305

> 2. There are only 103 members listed here, out of the 109 there should be. Might this be a similar scenario to the lower house where the others _are_ on the site, but not being returned from the search? (It might be useful to cross-check this list against the one on Wikipedia to track these down)

I've tried matching the people from Wikipedia's list to those from our scraper, and that let me find three more that are on the National Assembly site:

```
876,Sen.,"GEORGE SEKIBO","Rivers East","Rivers State",senate,"",Sen,"Peoples Democratic Party",PDP,http://www.nass.gov.ng/images/mps/876.jpg,2015,http://www.nass.gov.ng/mp/profile/876,"Rivers East, Rivers State",07063212133
91,Sen.,"LILIAN EKWUNIFE","Anambra Central","Anambra State",senate,"",Sen,"Peoples Democratic Party",PDP,http://www.nass.gov.ng/images/mps/91.jpg,2015,http://www.nass.gov.ng/mp/profile/91,"Anambra Central, Anambra State",""
937,Sen.,"OLAKA NWOGU","Rivers South East","Rivers State",senate,"",Sen,"Peoples Democratic Party",PDP,http://www.nass.gov.ng/images/mps/avatar.jpg,2015,http://www.nass.gov.ng/mp/profile/937,"Rivers South East, Rivers State",08033140757
```

There are three remaining that I can't find on the National Assembly site:

* Abdulsalami Ohiare (Kogi Central)
* Mao Ohuabunwa (Abia North)
* Air Marshal Isaac Alfa (Kogi East)

I don't think there's much we can do about that other than report it to the National Assembly site. So I think that means one Senator I know nothing about at the moment. 

I'll rebuild the data with those extra three, but I hope the others won't block merging this first version of the data anyway.

> 3. The one member of the Labour Party not reflected in the Wikipedia hemicycle seems correct. I'm assuming that graphic was created after the initial results were announced, before the Tribunal overturned this one (http://www.vanguardngr.com/2016/07/plans-confront-pdp-delta-omo-agege/). 

Ah, OK - thanks for figuring that out.

> 4. Skipping the elections for now is probably sensible, though it would be good to split these up on Wikidata at some point so we _can_ restrict the import to only Senate elections.

OK, noted.

> 5. Are you planning to map the Senators to Wikidata? It certainly doesn't need to be done in this PR, but it would be useful to at least point at an Issue to remind us that we'll want to do it at some point. 

Yes, that's a good idea - I've created a ticket for it here: https://github.com/everypolitician/everypolitician-data/issues/25307 and assigned it to myself.

tmtmtmtm:
The senator for Kogi Central seems to actually be Ahmed Ogembe (http://thenewsnigeria.com.ng/2017/02/sen-ogembe-decries-abductions-killings-in-kogi-central/), who's at http://nass.gov.ng/mp/profile/891

Mao Ohuabunwa is at http://nass.gov.ng/mp/profile/506

The senator for Kogi East is listed on the official site as being Abdul Abubakar — http://www.nass.gov.ng/mp/profile/520 — although it looks like he was replaced last month by Atai Aidoko http://nigeria.tvcnews.tv/2017/01/10/aidoko-replaces-isaac-alfa-as-kogi-east-senator/

mhl:
Thanks very much for finding them and investigating. Ahmed Ogembe was actually being scraped already, so we're one person short, still. However, I'd suggest that, based on the experience of doing this for the House of Representatives, the final missing Senator should become apparent at the slightly later stage where we're mapping the areas to MapIt IDs, and we could add them then.

mhl:
Here's an updated screenshot of viewer-sinatra's presentation of the party groupings based on the latest version of this pull request, with 5 extra people:

updated-senate-groupings

One is senator still missing, we hope to locate after doing more work on the area data. It would be good to merge this anyway.

The hemicycle is as @tmtmtmtm quoted before - the latest version as of writing is:

https://en.wikipedia.org/w/index.php?title=Senate_of_Nigeria&oldid=763995793
nigerian-senate-hemicycle

The discrepancy in party groupings was explained by @tmtmtmtm above:

> The one member of the Labour Party not reflected in the Wikipedia hemicycle seems correct. I'm assuming that graphic was created after the initial results were announced, before the Tribunal overturned this one (http://www.vanguardngr.com/2016/07/plans-confront-pdp-delta-omo-agege/).

data/Nigeria/Senate/sources/instructions.json
```diff
@@ -0,0 +1,26 @@
+{
+  "sources": [
+    {
+      "file": "morph/data.csv",
+      "create": {
+        "from": "morph",
+        "scraper": "everypolitician-scrapers/nigeria-national-assembly",
+        "query": "SELECT * FROM data WHERE js_position = 'Sen' AND area != ''"
```
tmtmtmtm: sorry for not spotting this earlier, but this gives a build warning as it doesn't include an `ORDER BY` (as it otherwise creates lots of spurious diffs when the source order changes later)

mhl: OK, I've changed that, regenerated the CSV file and regenerated `countries.json`. (`ORDER BY id` was sorting on the IDs as strings, both in my local sqlite DB and via morph's SQL API, so I made the `ORDER BY CAST(id AS int)`.)

I'll make a note to also add an `ORDER BY` to the query for `Nigeria/Representatives`.

tmtmtmtm:
:man_dancing:

data/Nigeria/Senate/sources/instructions.json
tmtmtmtm: FYI: I don't think the cast to int is actually necessary — the important part is just that the order is consistent over time. But there's certainly no harm in doing it.

PR #25071, opened 2017-02-03
Nigeria: Ignore scraped representatives with no area information

mhl:
The Nigerian National Assembly website has a blank constituency field
on some representative pages. These appear to be cases where there is a
duplicate page for the person that does have area information, or where
the election of that representative is in dispute. (I suspect that
what's happening is that people notice people who shouldn't appear in
the list of representatives on the site, and removing the constituency
data is the simplest way of stopping them appearing in that list. Or
something like that, anyway.)

Removing these cases stops some people appearing multiple times, and
means you can work with the data assuming that everyone has a valid
area.


tmtmtmtm:
:+1: — thanks @mhl

PR #25078, opened 2017-02-03
Add a script to ensure commits referenced in countries.json are valid

mhl:
The countries.json file provides URLs to Popolo JSON files and CSV files
which include the full object name of a commit that contains the
up-to-date version of that file.  It's very easy, however, to forget to
rebuild the countries.json file after updating the data for a countries,
or rewriting the history of a branch in which you made an update.

This script should detect both of those errors, emit a warning and exit
in error if any problems were found.

If this is useful, it could be added to checks run by Travis for each
pull request.

(This is just a proof-of-concept to see if people think it'd be
helpful, so I haven't written any tests yet.)

### DIFF
--- a/bin/check_countries_json_commits.rb
+++ b/bin/check_countries_json_commits.rb
@@ -1,0 +1,60 @@
+#!/usr/bin/env ruby
+
+# This script can catch errors where you've forgotten to rebuild
+# countries.json.  It'll emit a warning and exit in error if it can't
+# find a commit that's referenced in the countries.json file which
+# isn't in the history of the current branch, or if the contents of
+# the file that's referenced is different between the current version
+# and the referenced one.
+
+require 'everypolitician'
+require 'set'
+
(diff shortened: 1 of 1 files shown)


bin/check_countries_json_commits.rb
```diff
@@ -0,0 +1,60 @@
...
+known_commits = all_commits_in('HEAD').to_set
+
+success = true
+
+# For each country check the Popolo URL for the legislature and the
+# CSV URL for each legislature:
+
+EveryPolitician.countries.each do |c|
+  c.legislatures.each do |l|
+    success &&= check_url(l.popolo_url, known_commits)
```
mhl: n.b. this should also check that the `sha` attribute of the legislature is in known_commits (and that it's the same as the one in popolo_url)

mhl:
Note to self: see discussion in x-mark on the EP Slack for details of what to do next

PR #25088, opened 2017-02-03
Rows

marfi:
Proper rows count


tmtmtmtm:
Good catch — thanks @marfi

PR #25096, opened 2017-02-03
Cook Islands: Wikidata IDs for terms + reordering

marfi:
Smaller to bigger, plus wikidata_ids where available


tmtmtmtm:
Thanks @marfi — if you want to change the 14th as well, it's Q24895829

PR #25098, opened 2017-02-03
Philippines: Wikidata ids for terms

marfi:
Better than wikipedia url


tmtmtmtm:
Thanks @marfi — I've also added the 17th Congress, and tidied up a lot of the term information on Wikidata.

PR #25099, opened 2017-02-03
Sri Lanka: Wikidata id for term

marfi:
Better than wikipedia url


tmtmtmtm:
Thanks @marfi — I've also added some better data for this term and others back to the 9th on Wikidata

PR #25101, opened 2017-02-03
This was for previous term

marfi:


tmtmtmtm:
Thanks @marfi — I can't find any Wikidata pages yet for the individual terms in Venezuela. At some point I'll probably get around to adding them myself, and then we'll be able to change this to point at them.

PR #25487, opened 2017-02-10
Nigeria (Senate): Add Shine Your Eye IDs

struan:
This will allow adding the IDs from the existing ShineYourEye MP profiles.

```
  ⚥ data for 0; 0 added

Creating names.csv
Persons matched to Wikidata: 0 ✓ | 364 ✘
Parties matched to Wikidata: 6 ✓ 
Areas matched to Wikidata: 0 ✓ | 357 ✘
```


data/Nigeria/Senate/sources/instructions.json
```diff
@@ -21,6 +21,21 @@
         "from": "group-wikidata",
         "source": "manual/group_wikidata.csv"
       }
-    }
+    },
+    {
+      "file": "morph/shineyoureye.csv",
+      "create": {
+        "from": "morph",
+        "scraper": "struan/nigeria-shineyoureye",
```
tmtmtmtm: We have quite a strong preference not have to have new scrapers outside the `everypolitician-scrapers` account. Are you able to move this across?

data/Nigeria/Senate/sources/instructions.json
```diff
@@ -21,6 +21,21 @@
...
+        "query": "SELECT *, NULL as position FROM data WHERE position = 'Senator'"
```
tmtmtmtm: I think this will need to alias the `identifier__pombola_slug` column on the way through. At the minute that creates data in the format:

```json
      "identifiers": [
         {
           "identifier": "samuel-egwu",
           "scheme": "pombola_slug"
         }
       ],
```

But 'pombola_slug' doesn't really describe what the source is. I suspect this should just be `source__shineyoureye`.

mhl:
I've run the reconciliation myself (but using the fuller version of names we can now get from the scraper) and there was one match @struan included that I didn't:

```
core_person:427,2d19dc00-2a3d-48a4-9373-f05e3fec46c9
```

That's: http://www.shineyoureye.org/person/uche-lilian-ekwunife/ in Pombola, and LILIAN EKWUNIFE http://www.nass.gov.ng/images/mps/91.jpg from the National Assembly website, so I think @struan was right and somehow I missed her.

data/Nigeria/Senate/sources/instructions.json
mhl: I worry that just having `shineyoureye` may turn out to be confusing if / when we migrate to the static ShineYourEye site. Also, I'd like to include `slug` in the scheme, since it might well be useful to also include the primary key ID as well. So then you could have the two schemes:

* `source__shineyoureye__pombola_slug`
* `source__shineyoureye__pombola_id`

... perhaps?

tmtmtmtm: I don't think I understand deeply enough why this is being included here (or indeed what the two different values might be used for) and what the lifespan of the usefulness is. 

Possibly the most visible usage will be that once this is in the data, EveryPolitician.org will start displaying it on each card, and I'd like to be able to turn that into a link to the ShineYourEye site. If you're not maintaining URL-compatibility between the two versions of the site, then we'll need to tweak the identifier → URL mapping at that point, but that's fairly simple to do. If the prior identifier also becomes largely meaningless at that point, then I'm not sure it'll be worth continuing to include it.

I suspect you're also wanting to include it/them, though, so that it makes something else easier at your end?

mhl: The immediate use for us is so that we can make sure that the old URLs for people on the Pombola-based shineyoureye.org work on the new static site without any redirection.

I do care much more about this being merged soon than I do about the name of the identifier scheme, though, so feel free to ignore this.

tmtmtmtm: I'm still struggling to follow this a little. ISTM that the field that Pombola internally calls the slug is really the outward facing identifier, as it's what's used in the URLs, and if you're planning to also keep the URLs the same on the new site, then I think that this is definitely OK as a plain `identifier__shineyoureye`.

I _think_ that the internal Pombola IDs are a bit of a red herring here, though I'm open to persuasion that there's some value in EP having them. If there is, then I think we should come up with a different name for them to convey how someone would actually consume/use them.

mhl: The point where I was concerned about the confusion arising is if in the future we change the URLs for people on the new static ShineYourEye to some different scheme. Then we'd need to switch to getting those identifiers from a different source than the old Pombola site anyway, and changing the meaning of `source__shineyoureye` would confuse me more than deprecating one called `source__shineyoureye_pombola_slug` or `source__shineyoureye_pombola` and adding a new one called `source__shineyoureye_static` (or whatever) at that point. I can see that you might take the opposite point of view, though: that there'll only be one shineyoureye.org at a time, and so it's less confusing to just have one identifier scheme that always can be used to generate URLs to shineyoureye.org.

The internal Pombola IDs may indeed be a red herring - we don't have an immediate use for them, hence why I said "might well be useful" rather than "will be useful" :)

data/Nigeria/Senate/ep-popolo-v1.0.json
```diff
@@ -138,54 +256,141 @@
...
+          "note": "alternate"
+        }
+      ]
     },
     {
+      "birth_date": "1962-12-19",
       "contact_details": [
         {
+          "type": "email",
+          "value": "abs@sarakimail.com bukisaraki@aol.com"
```
tmtmtmtm: There are about 5 examples of email fields that are doubled up like this. In the scraper you can separate those with a semicolon to have them flow through as separate values here. (Or, I guess, at a pinch you could update the SELECT to replace on space, but it's probably better to do it in the scraper)

mhl:
I've updated this for master and the updated scraper data, and added an ORDER BY in the instructions for that source. The new senator doesn't appear to be in the Pombola ShineYourEye data.

(In the course of doing this I did notice *another* missing senator, but I'd like to deal with adding him separately afterwards - it's more urgent for us that we have most of this data present.)

PR #25710, opened 2017-02-13
Nigeria (Representatives): use ORDER BY to order data from Morph

mhl:
When extracting data from the Morph scraper, we weren't specifying an
order. This produces a warning when recreating that source if you've
removed it - the lines in sources/morph/data.csv should be predictably
sorted to minimize the diffs of that file.


mhl:
Oops, I put Senate instead of Representatives in the commit message - fixed now

mhl:
@tmtmtmtm Oops, yes - sorry for forgetting to note that. I generated it with:

```
sqlite3 -csv -header ~/ep-scrapers/nigeria-national-assembly/data.sqlite "SELECT * FROM data WHERE js_position = 'Hon' AND area != '' AND NOT name LIKE '%Anonymous%Anonymous%' ORDER BY CAST(id AS int)" | tail -n +2 > data/Nigeria/Representatives/sources/morph/data.csv
```

... where I'd populated that with data from scraper prior to the full name changes (which might account for the differences)

PR #26457, opened 2017-02-22
Error if Person source includes Membership fields

tmtmtmtm:
When building Person + Membership information, we need to be very careful
not to include "Membership"-type data (party, constituency, term, start/end
date) from any Person source, as that will get merged into every membership
that the Person has had. This appears safe if there's only a single term,
and no-one changes party or area during that, but the instant someone has a
second membership, bad things start to happen.

If we _do_ see any, when fetching the list of fields, raise an error.

This is still WIP whilst we fix all the countries.

Fixes https://github.com/everypolitician/everypolitician/issues/589


tmtmtmtm:
Current problems:

* [] Canada/Commons: represent-41.csv `[:group, :area, :group_id, :term]`
* [] Costa_Rica/Assembly: wikipedia.csv `[:area, :group]`
* [] Gibraltar/Parliament: wikipedia.csv `[:term, :group]` (Uses `term_match`)
* [] Isle_of_Man/House_of_Keys: images.csv `[:area, :term]`
* [] Italy/House: openpolis.csv`[:group, :area, :term]` (uses `term_match`)
* [] Spain/Congress: official.csv `[:group, :group_id, :term, :start_date, :end_date, :area]`)
* [] Zimbabwe/Assembly: official.csv `[:group, :area]`

Issue #25274, opened 2017-02-06
Turks and Caicos Islands (House of Assembly)


ondenman:
# Problem with Existing Data

## Legislature

Turks and Caicos Islands (House of Assembly)

## Problem

Lots of incoming data but no changes showing. New term needs adding.
https://github.com/everypolitician/everypolitician-data/pull/25271

ondenman:
It appears that the 2012 is missing a member.

The member appears in the Wikipedia [article](https://en.wikipedia.org/w/index.php?title=Turks_and_Caicos_Islands_general_election,_2012&oldid=755144382) for that election but is not in the [live data](http://everypolitician.org/turks-and-caicos-islands/assembly/term-table/2012.html).

ondenman:
Membership data for 2012 is coming in from the WP person source. It shouldn't. We should be getting membership info only from membership sources. See: https://github.com/everypolitician/everypolitician-data/pull/26457

ondenman:
Unwanted data is coming in from Wikipedia.
```
name,wikiname,party,area,term,id
Akierra M. Missick,Akierra Missick,PNP,Leeward,2012,akierra_m._missick
Amanda A. Misick,"",PNP,Cheshire Hall & Richmond Hill (results voided in February 2013),2012,amanda_a._misick
```

ondenman:
I've opened an exploratory PR: https://github.com/everypolitician/everypolitician-data/pull/26675. 

I think the best approach would be to sort out the existing term first: make Wikipedia the membership source and have person data come in from the archived file. Wikipedia gives us an extra member missing from the archive but we lose four others. I have looked in the WP article history for those members but can find no mention of them. They don't appear to be elected members, which would explain why.

I'm not sure how yet, but I believe these missing members should come from an upstream source. Though, they could be stored in an archive file.

ondenman:
I've opened a PR that splits the 2012 archive source into two. One for elected members (which will be pulled in from Wikipedia) and another for the appointed members.

ondenman:
In addition to elected members, the legislature has three appointed members are three ex officio members.

The wikipedia scraper captures only elected members.

![screen shot 2017-03-08 at 11 28 06](https://cloud.githubusercontent.com/assets/4107953/23702288/80407e28-03f2-11e7-8d39-a2142ee4e7ae.png)

ondenman:
[...]

PR #26565, opened 2017-02-23
Turks and Caicos (Assembly): Amend wikipedia query to import person data only from 2012

ondenman:
In a forthcoming PR, Wikipedia will be the source for term 2016 membership data. 2016 person data will come from the official source.

This is a step towards that.

Part of: https://github.com/everypolitician/everypolitician-data/issues/25274


ondenman:
Closing in favour of: https://github.com/everypolitician/everypolitician-data/pull/26968#event-979905022

Issue #25302, opened 2017-02-06
Alderney: broken import (new term)


ondenman:
# Problem with Incoming Data

## Legislature
Alderney (States)

## Problem
New names coming in but they need to be added to new term.

## Steps
- [x] Add test framework to scraper
- [x] Refactor scraped
~Decorate image paths~
~Change term id in scraper from 2014 to 2017~
- [x] Remove term name from scraper
- [x] Archive old term
- [x] Import data

ondenman:
I've removed the _decorate image paths_ step as it is not require for the data import. There is an issue for it in the scraper repo:https://github.com/everypolitician-scrapers/Alderney/issues/5

ondenman:
The legislature does not have distinct terms.

_"Elections in Alderney are held for the positions both of President and Member of the States of Alderney. The President of the States of Alderney is directly elected every four years. Half of the ten States Members are elected every two years for a four-year mandate."_ https://en.wikipedia.org/wiki/Elections_in_Alderney

Instead of renaming the term within the scraper, I am now removing it. The term can be set in the import query.

PR #26573, opened 2017-02-23
Alderney States: archive 2014 term

ondenman:
This is a step towards importing the next term. Part of: https://github.com/everypolitician/everypolitician-data/issues/25302

The legislature does not have set terms.

"Elections in Alderney are held for the positions both of President and Member of the States of Alderney. The President of the States of Alderney is directly elected every four years. Half of the ten States Members are elected every two years for a four-year mandate." [https://en.wikipedia.org/wiki/Elections_in_Alderney]

Each two year period is treated as an arbitrary term. There has been a recent election and the makeup of the legislature has changed.

The previous term data needs to be archived as the source site has been updated to reflect the recent changes.

__Note to merger:__

https://morph.io/tmtmtmtm/Alderney can now be deleted as the scraper has moved.


ondenman:
@tmtmtmtm Yes. I've updated the description.

PR #26675, opened 2017-02-24
WIP: Turks add caicos exploratory branch

ondenman:


ondenman:
Notes:
There is a new member coming in from Wikipedia but we're losing four others. I've looked through the article history but can find no mention of them. We could add them to an archive file but it would be best to add them upstream.

PR #28270, opened 2017-03-16
Complain about overlapping terms

tmtmtmtm:
If any terms in the source file overlap, blow up noisily.

Part of https://github.com/everypolitician/everypolitician/issues/594


tmtmtmtm:
The only legislature that currently breaks under this is `Czech_Republic/Deputies`

PR #28717, opened 2017-03-20
Travis: Don't send 'success' notifications to Slack

tmtmtmtm:
There's a lot of noise in the Travis channel on Slack, and we almost
never want to know when a build succeeded — only when it fails. So turn
off success notifications.

See: https://docs.travis-ci.com/user/notifications/#Changing-notification-frequency


tmtmtmtm:
Looks like this didn't work. @chrismytton any ideas?

PR #29233, opened 2017-03-23
[WIP] France (National Assembly): Add mapit area source

chrismytton:
This adds a new source type of `area-mapit`.

Current features/problems:

- [ ] `identifiers` and `other_names` are getting clobbered rather than merged. I think this is a problem in `rake_build/generate_ep_popolo.rb`
- [ ] The scraper I refer to in `instructions.json` doesn't exist yet
- [ ] `morph/mapit-areas.csv` and `reconciliation/mapit-areas.csv` were created manually and only have one entry
- [ ] The `Source::Mapit` class is a copy/paste of `Source::Area`, all I've done is replace "wikidata" with "mapit"


chrismytton:
@tmtmtmtm Does this look like it's heading in _roughly_ the right direction?

tmtmtmtm:
> I think this is a problem in rake_build/generate_ep_popolo.rb

Yes — the code you've added tells it to simply clobber everything other than `id`:

```
          existing.merge!(area) do |key, old, new|
            key == :id ? old : new
          end
```

at this point `existing` is:

```json
{:id=>"65_1",
 :name=>"Hautes-Pyrénées, 1",
 :type=>"constituency",
 :identifiers=>[{:identifier=>"2802223", :scheme=>"openstreetmap"}, {:identifier=>"Q3402121", :scheme=>"wikidata"}],
 :other_names=>[{:lang=>"fr", :name=>"première circonscription des Hautes-Pyrénées", :note=>"multilingual"}]}
```

and you've generated an `area` of:

```json
{:id=>"65_1", :identifiers=>[{:identifier=>"1189872", :scheme=>"mapit"}], :other_names=>[]}
```

So your addition to `generate_ep_popolo` will probably want to do something cleverer. (Though that should probably be shared with the Wikidata one, so that it doesn't matter which order these come in)

(Note that for a good test case here, you probably want your mocked up result from the scraper to actually have at least one `other_name` as well…)

PR #29363, opened 2017-03-24
Ghana Parliament: rename source files

ondenman:
everypolitician/everypolitician#563 + everypolitician/everypolitician#561


data/Ghana/Parliament/sources/instructions.json
```diff
@@ -11,7 +11,7 @@
       "type": "membership"
     },
     {
-      "file": "manual/members-term-6.csv",
+      "file": "manual/official-term-6.csv",
```
tmtmtmtm: If you're going to move this, I suspect you should be moving it to an `archive/` directory

PR #29433, opened 2017-03-24
Switch to deep_merge for merging areas

chrismytton:
As part of adding Mapit info to France we want to switch to doing a deep merge of areas so that we can combine `identifiers` and `other_names`.

This change switches to using the `deep_merge` gem which has mean un-reconciling some of the France Wikidata ids because it was incorrectly including historic Wikidata entries in the Popolo.

Step towards https://github.com/everypolitician/everypolitician-data/pull/29233

# Notes to reviewer

I've run this across all legislatures and there was no diff produced, so this has the same behaviour as the previous code.


Gemfile
```diff
@@ -27,6 +27,7 @@ gem 'everypolitician-pull_request', github: 'everypolitician/everypolitician-pul
 gem 'everypolitician-dataview-terms', github: 'everypolitician/everypolitician-dataview-terms'
 gem 'field_serializer', github: 'everypolitician/field_serializer'
 gem 'csvlint'
+gem 'deep_merge'
```
tmtmtmtm: I was a little surprised that `rubocop` pick up on this, but it seems we've never actually turned it on for this repo (presumably because of how many errors we currently have!)

This isn't a particularly serious case, so I think we can live with it for now, but it would probably be worth trying to turn rubocop on fairly soon, even with a fairly large TODO file, so we at least start moving in the right direction.

PR #29437, opened 2017-03-24
Suriname (Assembly): rename source files

ondenman:
everypolitician/everypolitician#563


data/Suriname/Assembly/sources/instructions.json
```diff
@@ -1,13 +1,13 @@
 {
   "sources": [
     {
-      "file": "manual/members-2010.csv",
+      "file": "manual/official-2010.csv",
```
tmtmtmtm: If you're renaming this file anyway, you should probably move it to `archive`

Issue #25282, opened 2017-02-06
Guatemala (Congress) -- broken import


ondenman:
# Problem with Existing Data

## Legislature

Guatemala (Congress)

## Problem

Lots of membership changes. Statement count down. Party names changing.

ondenman:
Description updated.

ondenman:
The scraper is only picking up term 8.

ondenman:
158 members are being removed from term 7.
158 members are being added to term 8.
162 members are being removed from term 7.

ondenman:
The term is hardcoded in the scraper. Term 7 data no longer exists on morph.

ondenman:
Many members have an incorrect birth date:
```
325,Alsider Antonio Arias Rodríguez,FCN,FCN,8,http://www.congreso.gob.gt/manager/images/C3FB6058-38D8-7A18-FC66-797FC6D77297.jpg,Chimaltenango,2016-01-19,,congreso@congreso.gob.gt,http://www.congreso.gob.gt/perfiles.php?id=325
326,Héctor Melvyn Caná Rivera,UNE,UNE,8,http://www.congreso.gob.gt/manager/images/C86388D4-51F5-6988-6F58-798C92F28DDC.jpg,Chimaltenango,2016-01-19,,congreso@congreso.gob.gt,http://www.congreso.gob.gt/perfiles.php?id=326
327,Víctor Manuel Estrada Orellana,PP,Reformador,8,http://www.congreso.gob.gt/manager/images/EA3D3649-2E56-C564-B1AB-7994971286BF.jpg,El Progreso,2016-01-19,,congreso@congreso.gob.gt,http://www.congreso.gob.gt/perfiles.php?id=327
```

ondenman:
Email and area columns are empty in latest data.

ondenman:
There are fewer term 8 rows in the latest incoming data than term 8 rows in the live data. It's possible that some term 8 members need to be added to the archive.

ondenman:
I've archived term 7 members in: https://github.com/everypolitician/everypolitician-data/pull/30356

PR #30356, opened 2017-03-29
Guatemala (Congress): Archive term 7 memberships

ondenman:
Part of: https://github.com/everypolitician/everypolitician-data/issues/25282

The [source site](http://www.congreso.gob.gt/legislaturas.php) lists members of the current term only. The scraper now clears the database on each run. As such, we're losing members of the previous term.

This PR extracts previous term memberships from the source file and stores them in the archive directory.


ondenman:
@tmtmtmtm I've updated the issue description.

tmtmtmtm:
@ondenman most of the other things here seem like they could be done in parallel with this, rather than queuing up behind it. Have you started the scraper work, for example?

PR #30686, opened 2017-03-31
Only deploy when countries.json changes

chrismytton:
We don't want to deploy everypolitician.org unless there has been a change to `countries.json`.

This uses the "TRAVIS_COMMIT_RANGE" environment variable to check if there has been a change to `countries.json` in the commit being built. If not then the deploy is skipped to avoid unnecessary spidering.

Follows on from https://github.com/everypolitician/everypolitician-data/pull/30681

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -41,7 +41,10 @@ deploy_viewer_static() {
 
 main() {
   if [[ "$TRAVIS_PULL_REQUEST" == false && "$TRAVIS_BRANCH" == master ]]; then
-    git diff --name-only $TRAVIS_COMMIT_RANGE -- countries.json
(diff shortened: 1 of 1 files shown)


tmtmtmtm:
I'm going to merge this anyway, as it'll cut out a lot of rebuilds when I go on a merge-spree, but it'd be good to fix this up anyway (assuming I'm not misreading it)

chrismytton:
> I'm not completely sure that I'm following this correctly, but isn't this saying "Skip if there are any files that aren't countries.json" — i.e. if a single commit updated countries.json and another file, this would report that there were no changes to countries.json?

Ah yep, you're absolutely right! I'll open another pull request to reinstate the ` -- countries.json` bit of the diff command that I (wrongly!) dropped.

PR #31019, opened 2017-04-03
Saint Kitts and Nevis (Assembly): Remover superfluous souce file

ondenman:
The file is no longer referenced in the import instructions.


tmtmtmtm:
👍

Issue #24210, opened 2017-01-12
Pakistan National Assembly -- Duplicate member in data


ondenman:
# Problem with Existing Data

A member is listed twice within the same term.

![screen shot 2017-01-12 at 16 55 47](https://cloud.githubusercontent.com/assets/4107953/21899438/115d48cc-d8e8-11e6-816d-538c1f14f14f.png)

http://everypolitician.org/pakistan/assembly/term-table/14.html

## Legislature

Pakistan National Assembly 

## Proposed Solution

Map the member's upstream ID with their original UUID. Do not archive the data associate with the member's original ID.

PR #31195, opened 2017-04-04
Pakistan (Assembly): Add end date column to archive data

ondenman:
This is a step towards fixing: https://github.com/everypolitician/everypolitician-data/issues/24210

One member has two memberships in the current term. There is no end date for the first membership and so it overlaps the second.

The first membership is stored in the archive directory. As a step towards adding the end date for this membership, this PR adds an `end_date` column to the archived membership file.


ondenman:
@tmtmtmtm I've amended the commit message and force pushed.

PR #32059, opened 2017-04-11
Thailand Assembly: switch away from personal scraper on morph.io

davewhiteland:
Moved scraper from https://morph.io/davewhiteland/thailand-national-assembly to https://morph.io/everypolitician-scrapers/thailand-national-assembly

There's no legacy data up there: the CSV being generated is identical in both cases, so this should be a seamless transfer.

Also set rebuilder webhook on the new scraper.


davewhiteland:
Noted:  see #32108 — that PR will hopefully deal with the data changes, so when that's done, this one should then be OK

PR #32261, opened 2017-04-13
Give Parties/Groups UUIDs

tmtmtmtm:
As a first step towards giving every Group a UUID, we take the very small
step of looking for an groupidmap file (in a known location, similar to how
the person version works), and if the group_id from the source has a
mapping in that, we swap to that.

For now we simply ignore all other cases, and continue to use whatever
group_id we have. There are no tools yet to create these files — they'll
all need to be done by hand initially. We can evolve those later when we
discover the issues around them. (We also need to remember to fix the
group_wikidata file. I had considered simply using the Wikidata ID as our
ID, but they're not necessarily stable.)

If this approach works, we possibly want to change how person-UUID mapping
also works, so that they're more harmonised.

part of https://github.com/everypolitician/everypolitician/issues/428


tmtmtmtm:
Need to consider whether we need to preserve the original ID as a legacy identifier or the like, similar to what we did with People.

PR #33799, opened 2017-04-25
Update version of Rubcop

tmtmtmtm:
Update to the latest rubocop and autoclean the Gemfile and the
per-legislature rakefiles. Other cleanups can come later.


tmtmtmtm:
I accidentally committed some rake tasks that I only had locally, but we may as well also have them here. (I should really open new PRs for each of them, but in the interests of expediency, I'll just merge them from here)

PR #43201, opened 2017-07-07
Benin Assembly: Amend term id

ondenman:
The current term was incorrectly named the 6th when in fact it is the 7th. The term name was updated in a previous PR. The scraper has since been updated to record the current term's ID as 7.

The scraper location will need to be moved to ep-scrapers. This PR updated the terms file with the correct term id and imports the term id as 7.


ondenman:
I'm not sure this has gone quite right as there are more members being added as those being removed.

PR #48615, opened 2017-08-03
Refresh countries.json

chrismytton:
This was produced by running `bundle exec rake countries.json` in the
root of the repo.

This seems like quite a big diff, which may indicate an underlying issue
with how countries.json is or isn't getting updated.


tmtmtmtm:
superseded

PR #53025, opened 2017-09-04
Pakistan (National Assembly): refresh official

everypoliticianbot:
Automated refresh of official for Pakistan - National Assembly

#### Output

```
Switched to a new branch 'pakistan-assembly-1504539673'
Add memberships from sources/morph/official.csv
1057 -> 0cb478ad-d3ab-40f8-b1d0-75c222f68df1
Add memberships from sources/archive/official-term-14.csv
Merging with sources/morph/wikidata.csv
Data Mismatches
* 16 of 363 unmatched
	{:id=>"Q134068", :name=>"Nawaz Sharif"}
	{:id=>"Q26924336", :name=>"Muhammad Ibrahim Jatoi"}
	{:id=>"Q18763831", :name=>"Ashiq Husain Khan"}
	{:id=>"Q18816019", :name=>"Rai Mansab Ali Khan"}
	{:id=>"Q7636919", :name=>"Sumaira Malik"}
	{:id=>"Q26924450", :name=>"Zulfiqar Ali Behan"}
	{:id=>"Q18763842", :name=>"Ayesha Raza Farooq"}
	{:id=>"Q17287330", :name=>"Tahira Asif"}
	{:id=>"Q33076529", :name=>"Usman Badini"}
	{:id=>"Q3073499", :name=>"Javed Hashmi Baghi"}
Merging with sources/morph/genderbalance.csv
Data Mismatches
* 2 of 344 unmatched
	{}

Top identifiers:
  347 x wikidata
  15 x freebase
  6 x viaf
  2 x fast
  2 x lcauth

Creating names.csv
  ☇ No dates for Ch. Pervez Elahi (Q2961554) as Deputy Prime Minister of Pakistan
  ☇ No dates for Sheikh Rasheed Ahmed (Q7462429) as Minister of Information and Broadcasting
Persons matched to Wikidata: 347 ✓ | 1 ✘
  No wikidata: Engineer Mohammad Usman Badaini (0cb478ad-d3ab-40f8-b1d0-75c222f68df1)
Parties matched to Wikidata: 19 ✓ 
Areas matched to Wikidata: 270 ✓ 
[pakistan-assembly-1504539673 81157ce] Pakistan: Refresh from upstream changes
 8 files changed, 75 insertions(+), 728 deletions(-)
 rewrite data/Pakistan/Assembly/unstable/positions.csv (68%)

```

PR #53037, opened 2017-09-04
Pakistan (National Assembly): Fix official site import

chrismytton:
This fixes the following issues with the official site import:

- 2 people have dissappeared from the official site, I've archived those off
- 1 [new person](http://na.gov.pk/en/profile.php?uid=1057) has appeared on the official site, I've reconciled them to Wikidata
- The new [Prime Minister has "Prime Minister of Pakistan" in his name](http://na.gov.pk/en/profile.php?uid=720), so I've fixed that upstream in https://github.com/everypolitician-scrapers/pakistan-national-assembly/commit/be0bec51c950842e5335726b7fc5d5d83d766aa8
- Fix photo URLs, upstream fix was https://github.com/everypolitician-scrapers/pakistan-national-assembly/commit/51abd21b6be62e6ae0e976b20960adbe26c7aff0

Because [the scraper for the official site](https://morph.io/tmtmtmtm/pakistan-national-assembly) isn't in the everypolitician-scraper account I can't manually trigger a run, so instead I've run the scraper locally and then manually generated a CSV from it and committed that.

Part of https://github.com/everypolitician/everypolitician/issues/612


chrismytton:
Replaces https://github.com/everypolitician/everypolitician-data/pull/53025

data/Pakistan/Assembly/sources/idmap/official.csv
```diff
@@ -349,3 +347,4 @@ id,uuid
 1053,c03a2279-1684-479a-8165-c6098ebb1174
 1054,5ed669b9-d426-4b19-8298-53cc6598d9e9
 1056,f67e27f4-7419-425c-879b-c89672659464
+1057,47223c4f-9ca3-4d33-928a-258adee23f2d
```
tmtmtmtm: This seems to have slipped into an earlier commit. It's not really part of archiving off the vanishing members.

data/Pakistan/Assembly/ep-popolo-v1.0.json
```diff
@@ -2958,10 +2954,10 @@
           "scheme": "wikidata"
         }
       ],
-      "image": "http://www.na.gov.pk/uploads/images/NA-300%20sMayryam.jpg",
+      "image": "http://www.na.gov.pk/uploads/images/final%20s01.JPG",
```
tmtmtmtm: Not quite sure what's going on here, but this image doesn't exist.

data/Pakistan/Assembly/ep-popolo-v1.0.json
```diff
@@ -4420,6 +4416,37 @@
...
+          "url": "http://www.na.gov.pk/uploads/images/NA-260%20sEngr.%20sMuhammad%20sUsman%20sBadini.jpg"
+        }
+      ],
+      "links": [
+        {
+          "note": "Wikipedia (en)",
+          "url": "https://en.wikipedia.org/wiki/Usman_Badini"
+        }
+      ],
+      "name": "Engineer Mohammad Usman Badaini",
```
tmtmtmtm: Hmmm. We probably want to start trapping prefixes like this. It's a pre-existing problem, so it doesn't need to be done in this change, but we'd probably want to fix it before letting any new people in with the same problem, especially as this is now a well-solved issue with a solution that can copied easily from other scrapers. It also seems like the sort of thing that's going to get in the way of our Wikidata prompts, so that might also accelerate us wanting to fix it.

data/Pakistan/Assembly/term-14.csv
```diff
@@ -191,7 +192,7 @@ cc6f31bb-70cf-4517-824b-981b15d9efa6,Muhammad Junaid Anwar Chaudhary,Muhammad Ju
...
 aa87aa59-451e-48a3-b755-15becf80df53,Muhammad Moeen Wattoo,Muhammad Moeen Wattoo,,,,PML(N),pmln,area/na-147_(okara-v),NA-147 (Okara-V),National Assembly,14,,,http://www.na.gov.pk/uploads/images/147.JPG,male,Q18815998,Q799577,Q12856786
 0e2c4697-c9b4-4df4-a90a-98c7af328ed5,Muhammad Muzammil Qureshi,Muhammad Muzammil Qureshi,,,,MQM,mqm,area/na-253_(karachi-xv),NA-253 (Karachi-XV),National Assembly,14,,,http://www.na.gov.pk/uploads/images/253.jpg,male,Q19517997,Q1265113,Q12857475
-7304486c-f12d-4009-b2db-ee47ec4f5771,"Muhammad Nawaz Sharif, Prime Minister Islamic Republic of Pakistan","Muhammad Nawaz Sharif, Prime Minister Islamic Republic of Pakistan",,,,PML(N),pmln,area/na-120_(lahore-iii),NA-120 (Lahore-III),National Assembly,14,,,http://www.na.gov.pk/uploads/images/120.jpg,male,Q134068,Q799577,Q12780067
+7304486c-f12d-4009-b2db-ee47ec4f5771,"Muhammad Nawaz Sharif, Prime Minister Islamic Republic of Pakistan","Muhammad Nawaz Sharif, Prime Minister Islamic Republic of Pakistan",,,,PML(N),pmln,area/na-120_(lahore-iii),NA-120 (Lahore-III),National Assembly,14,,2017-07-28,http://www.na.gov.pk/uploads/images/120.jpg,male,Q134068,Q799577,Q12780067
```
tmtmtmtm: If we're removing the Prime Minister suffix from names (which is definitely a good idea!), we should probably remove it from here too.

data/Pakistan/Assembly/sources/archive/official-term-14.csv
```diff
@@ -8,3 +8,5 @@ id,name,patronymic_name,address,phone,constituency,province,party,start_date,ima
...
 922,Abdul Hakeem Baloch,Abdullah,"Dur Muhammad Goth Dersano Channa, Malir.","03005250789, a.hakeembaloch@hotmail.com",NA-258 (Karachi-XX),Sindh,PML(N),2013-06-01,http://www.na.gov.pk/uploads/images/258.jpg,14,http://www.na.gov.pk/en/profile.php?uid=922,pmln,,2016-05-26
 967,Mrs. Alizeh Iqbal Haider,D/o Syed Iqbal Haider,"D-25, Block-IV, Clifton, Karachi",,,Sindh,PPPP,2013-06-01,http://www.na.gov.pk/uploads/images/313.jpg,14,http://www.na.gov.pk/en/profile.php?uid=967,pppp,,
+788,"Muhammad Nawaz Sharif, Prime Minister Islamic Republic of Pakistan",Mian Muhammad Sharif,"Shamim Farm, Jati Umra, Raiwind Road Lahore.","",NA-120 (Lahore-III),Punjab,PML(N),2013-06-01,http://www.na.gov.pk/uploads/images/120.jpg,14,http://www.na.gov.pk/en/profile.php?uid=788,pmln,male,2017-07-28
+924,Abdul Rahim Mandokhail,Abdul Rehman Mandokhail,"House No. 196 Tehsil Road, Zhob",0346-5303971,NA-260 (QUETTA-CUM-CHANGAI-CUM-NUSHKI (OLD QUETTA-CUM-CHAGAI-CUM-MUSTANG)),Balochistan,PMAP,2013-06-01,http://www.na.gov.pk/uploads/images/260.jpg,14,http://www.na.gov.pk/en/profile.php?uid=924,pmap,,2017-05-20
```
tmtmtmtm: These end dates don't seem to have been in the original file, so I'm assuming you found them from an external source? It would be useful to credit that, if so.

data/Pakistan/Assembly/ep-popolo-v1.0.json
chrismytton: This turned out to be a problem in the scraper which was causing lots of images to break. Fixed in https://github.com/everypolitician-scrapers/pakistan-national-assembly/commit/51abd21b6be62e6ae0e976b20960adbe26c7aff0.

data/Pakistan/Assembly/sources/idmap/official.csv
chrismytton: Oops, I've tidied up that commit so it only includes archive related things now.

data/Pakistan/Assembly/term-14.csv
chrismytton: Good point, I've removed that from the archived source.

data/Pakistan/Assembly/sources/archive/official-term-14.csv
chrismytton: I've added a link to the source to the commit message 👍

data/Pakistan/Assembly/ep-popolo-v1.0.json
chrismytton: I've opened https://github.com/everypolitician/everypolitician-data/issues/53128 to capture the problem with the name prefixes.

chrismytton:
@tmtmtmtm I've addressed you comments, so I think this is ready for review again.

chrismytton:
The scraper has now run, so I've updated this pull request to use the output from morph, rather than the manually generated version.

PR #57295, opened 2017-10-06
Argentina (Senado): update the Wikidata IDs for parties

mhl:
I'm somewhat confused about whether we're getting the distinction
between parties and parliamentary alliances right here - the official
source I think mostly lists alliances, but this may in some cases link
them to parties with very similar names to the alliances. This is
probably a good basis to start from, however.


tmtmtmtm:
Thanks @mhl

PR #85337, opened 2018-03-16
Fix image URLs from the old Pombola-based ShineYourEye

mhl:
The old Pombola-based ShineYourEye has moved from www.shineyoureye.org
to pombola.shineyoureye.org, which broke lots of image URLs for Nigerian
policitians. This pull request fixes those images URLs and brings in a
couple of incidental changes (one email address added and one removed).


tmtmtmtm:
superseded by #85333 and #85334

PR #114048, opened 2018-09-27
Use a safe downcase fuction for ruby 2.3 and 2.4

tmtmtmtm:
Ruby 2.4 changed String#downcase to work on all Unicode strings, rather
than only ASCII ones. This causes us some difficulty as it makes it
unstable to generate IDs, and to sort.

In order to migrate to 2.4 we need to make everything work consistently
across both versions.

To assist with this we introduce a String#to_lower function, which will
work like the ruby 2.4 downcase even under 2.3

This will allow us to first migrate all the data to this new version, and
then do the wider code upgrade to 2.4

Part of https://github.com/everypolitician/everypolitician/issues/614


tmtmtmtm:
This didn't work out well, so instead I updated to 2.4 in https://github.com/everypolitician/everypolitician-data/pull/135293 and then manually fixed all IDs that changed.

PR #115700, opened 2018-10-11
Include everything from terms.csv in the Popolo

tmtmtmtm:
Rather than only creating Events for terms that have members, include all
the entries from terms.csv.

We still need to make sure, however, that we report on any memberships of
terms that don't exist.

Closes https://github.com/everypolitician/everypolitician/issues/642


tmtmtmtm:
We probably need to give a little more thought to how we report the number of terms in `unstable/stats.json` — we only previously had one concept here, but now there are two: how many Terms we know about and how many Terms we have members for.

PR #119770, opened 2018-11-09
remove all OCD-handling code

tmtmtmtm:
We no longer use OCD division IDs in any significant way. Some scrapers
still provide them as IDs, but we are essentially treating those as opaque
IDs with no inherent meaning, rather than deriving any other data from
them. All legislatures which previously used this code now get the relevant
information from Wikidata instead.


tmtmtmtm:
I have verified that all legislatures still build cleanly.

PR #119823, opened 2018-11-09
rename 'type' instruction to 'generates'

tmtmtmtm:
Make it clearer what this field represents.

Closes https://github.com/everypolitician/everypolitician/issues/325


tmtmtmtm:
I'm not sure that this is really an improvement.

PR #152060, opened 2019-04-23
Revert "South Africa (National Assembly): refresh wikidata"

tmtmtmtm:
Reverts everypolitician/everypolitician-data#151848 — this lost lots of election data.


tmtmtmtm:
@jacksonj04 I'm not sure what you meant by "improved handling of elections with no data" on the original PR — this is actually losing a lot of useful election data. From what I can see it's because there's a lot of mismatches between data for the parliamentary elections and the presidential ones. For example, https://www.wikidata.org/wiki/Q591787 contains lots of information that should really be on https://www.wikidata.org/wiki/Q3586539, largely because different language Wikipedias have articles that don't quite align: in many cases there will really need to be at least 3 separate items; one for the presidential election, one for the parliamentary election, and one for the general election that includes both (and potentially also local elections etc.). EP will pull in anything that with an 'office contested' of Member of the Parliament of South Africa (Q6814221), and these are going missing because they don't have that (most have either the President or the Prime Minister instead.)

Unless/until these elections have that, you shouldn't merge any changes to the `elections.json` file that lose data like this. I can think of four possible options here:

1. Any time there's a PR updating all of the 'wikidata' files, manually revert the elections.json file before merging it. This is fairly clean, and not too much work, but it's likely to be annoying to do regularly, and obviously means that the elections info will drift gradually out of sync.
2. Switch the scrapers to more precisely target their webhook triggers, so that only individual sources are being rebuilt at once (and thus not rebuilding elections when the P39s or bio sources change). 
3. Fix up all the relevant elections to this body in Wikidata. This is obviously ideal, but is probably the most work. Once that's done, everything should flow cleanly, however.
4. Bulk add P541:Q6814221 to all of these. That will leave us with some items that aren't quite for the correct thing, but those can be then be separated out individually over time. This is a quick one-time cost, and will get everything working again at this end, but is arguably making Wikidata worse rather than better.

I think I'd probably go for either 2 or 4, or possibly both. I think Wikidata is already quite problematic here, with multiple Bonnie and Clyde items containing combined data, even where there are already individual Bonnie and/or Clyde articles. Adding extra data to the combined item will help make it clearer where there's duplication, and potentially easier to later separate them out, so in general I don't think it's going to be particularly problematic, as long as we're careful not to add info in cases where the data is already properly separated out (e.g. skipping anything with a 'has part' to another election). Other people may disagree with that approach, though. 

I'm also not sure what the impact of splitting the rebuilds would be on your normal workflow. In general, though, most legislatures have moved to having multiple small, single-source, rebuilds, rather than anything that rebuilds more than one source at once, as it's much easier to see what's going on, and prevents exactly this sort of issue, where problems with one source impact on otherwise clean changes from another one.

